PR #3, opened 2015-01-09
Generalize type signature of KV.of

kwestbrooks:
I found myself having to write this:
KV&lt;String, Iterable&lt;String&gt;&gt; kv = KV.&lt;String, Iterable&lt;String&gt;&gt;of("foo", new ArrayList&lt;String&gt;());

When I would have preferred to write this:
KV&lt;String, Iterable&lt;String&gt;&gt; kv = KV.of("foo", new ArrayList&lt;String&gt;());

What do you think about this change?


davorbonaci:
Thanks for the contribution. Looks like an improvement. Merged.

davorbonaci:
This pull request was reverted due to a number of non-trivial build breaks. We may reconsider this in the future.

PR #5, opened 2015-01-28
Introduces Proto2Coder.

perezd:
Introduces proto2 coder support for Cloud Dataflow.

### DIFF
--- a/sdk/src/main/java/com/google/cloud/dataflow/sdk/coders/Proto2Coder.java
+++ b/sdk/src/main/java/com/google/cloud/dataflow/sdk/coders/Proto2Coder.java
@@ -1,0 +1,139 @@
+/*
+ * Copyright (C) 2015 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.google.cloud.dataflow.sdk.coders;
+
+import com.google.protobuf.ExtensionRegistry;
+import com.google.protobuf.Message;
+import com.google.protobuf.Parser;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.lang.reflect.InvocationTargetException;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * An encoder using Google Protocol Buffers 2 binary format.
+ * <p>
+ * To learn more about Protocol Buffers, visit:
+ * <a href="https://developers.google.com/protocol-buffers">https://developers.google.com/protocol-buffers</a>
+ * <p>
+ * To use, specify the {@code Coder} type on a PCollection:
+ * <pre>
+ * {@code
+ * PCollection<MyProto.Message> records =
+ *     input.apply(...)
+ *          .setCoder(Proto2Coder.of(MyProto.Message.class));
+ * }
+ * </pre>
+ * <p>
+ * Custom message extensions are also supported, but the coder must be made
+ * aware of them explicitly:
+ * <pre>
+ * {@code
+ * PCollection<MyProto.Message> records =
+ *     input.apply(...)
+ *          .setCoder(Proto2Coder.of(MyProto.Message.class)
+ *          .withExtensionsFrom(MyProto.class));
+ * }
+ * </pre>
+ *
+ * @param <T> the type of elements handled by this coder, must extend {@code Message}
+ */
+public class Proto2Coder<T extends Message> extends CustomCoder<T> {
+
+  /**
+   * Produces a new Proto2Coder instance, for a given Protobuf message class.
+   */
+  public static <T extends Message> Proto2Coder<T> of(Class<T> protoMessageClass) {
+    return new Proto2Coder<>(protoMessageClass);
+  }
+
+  private final Class<?> protoMessageClass;
+  private final List<Class<?>> extensionClassList = new ArrayList<>();
+  private transient Parser<T> parser;
+  private transient ExtensionRegistry extensionRegistry;
+
+  Proto2Coder(Class<T> protoMessageClass) {
+    this.protoMessageClass = protoMessageClass;
+  }
+
+  /**
+   * Adds custom Protobuf extensions to the coder.
+   *
+   * @param extensionHosts must be a class that defines a static
+   *      method name {@code registerAllExtensions}
+   */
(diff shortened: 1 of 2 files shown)


davorbonaci:
Thanks Derek for this contribution.

We'll get back to you with specific code review comments as soon as we figure out how to best deal with the generated code.

perezd:
If you like, we could just inline it into the test. Its like 1300 lines, though :(

davorbonaci:
Let's not inline it.

We are currently investigating a couple of different options, such as 'maven-protoc-plugin', or separating the generated code into a separate artifact that can be pulled from Maven Central as a dependency.

perezd:
OK, makes sense.

sdk/src/main/java/com/google/cloud/dataflow/sdk/coders/Proto2Coder.java
```diff
@@ -0,0 +1,107 @@
...
+    private final ExtensionRegistry extensionRegistry = ExtensionRegistry.newInstance();
+
+    protected Proto2Coder(Class<T> protoMessageClass) {
+      this.protoMessageClass = protoMessageClass;
+    }
+
+  /**
+   * Registers Protocol Buffer message extensions with the coder's extension registry.
+   *
+   * @param extensions a variadic list of message extensions the coder uses properly parse messages
```
lukecwik: Missing "to":
the coder uses properly parse messages -> the coder uses to properly parse messages

perezd:
Fixed.

sdk/src/main/java/com/google/cloud/dataflow/sdk/coders/Proto2Coder.java
```diff
@@ -0,0 +1,107 @@
...
+ * @param <T> the type of elements handled by this coder, must extend {@code Message}
+ */
+public class Proto2Coder<T extends Message> extends StandardCoder<T> {
+
+    public static <T extends Message> Proto2Coder<T> of(Class<T> protoMessageClass) {
+      return new Proto2Coder<>(protoMessageClass);
+    }
+
+    private final Class<T> protoMessageClass;
+    private final ExtensionRegistry extensionRegistry = ExtensionRegistry.newInstance();
```
lukecwik: StandardCoder needs to support serialization to/from CloudObjects via the overridable asCloudObject() and the @JsonCreator annotated method of(...)

See for an example: 
https://github.com/GoogleCloudPlatform/DataflowJavaSDK/blob/master/sdk/src/main/java/com/google/cloud/dataflow/sdk/coders/AvroCoder.java#L118
https://github.com/GoogleCloudPlatform/DataflowJavaSDK/blob/master/sdk/src/main/java/com/google/cloud/dataflow/sdk/coders/AvroCoder.java#L165

Also, the protoMessageClass and extensionRegistry are not portable beyond Java.
To make this coder portable you will need to provide a cloud object mapping which is able to reconstruct the FileDescriptor. I would suggest the following format:
```json
{
  "message_type": "*the descriptors name*",
  "file_descriptor_proto": "*the file descriptor proto serialized*",
  "dependencies": [
     {
       "file_descriptor_proto": "*dependency file descriptor proto*",
       "dependencies": [ recursively more dependencies ],
     }
     ... more dependencies ...
  ]
}
```

You will then be able to reconstruct the FileDescriptor proto using:
https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/Descriptors.FileDescriptor#buildFrom(com.google.protobuf.DescriptorProtos.FileDescriptorProto, com.google.protobuf.Descriptors.FileDescriptor[])
And then get back the Descriptor using:
https://developers.google.com/protocol-buffers/docs/reference/java/com/google/protobuf/Descriptors.FileDescriptor#findMessageTypeByName(java.lang.String)

From then on you should be able to use the Descriptor to create messages.

sdk/src/main/java/com/google/cloud/dataflow/sdk/coders/Proto2Coder.java
```diff
@@ -0,0 +1,107 @@
...
+    @Override
+    public void encode(T value, OutputStream outStream, Context context) throws IOException {
+      value.writeTo(outStream);
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public T decode(InputStream inStream, Context context) throws IOException {
+      try {
+        T messageInstance = ((T) protoMessageClass.getMethod("getDefaultInstance").invoke(null));
```
lukecwik: Instead of using reflection, please store and use the parser for the messages Descriptor.

perezd: What do you propose the usage API is for that? Curious what the call site would look like.

perezd:
OK, I think we're ready for a review here.

sdk/src/main/java/com/google/cloud/dataflow/sdk/coders/Proto2Coder.java
```diff
@@ -0,0 +1,139 @@
...
+  }
+
+  /**
+   * Adds custom Protobuf extensions to the coder.
+   *
+   * @param extensionHosts must be a class that defines a static
+   *      method name {@code registerAllExtensions}
+   */
+  public Proto2Coder<T> withExtensionsFrom(Class<?>... extensionHosts) {
+    for (Class<?> extensionHost : extensionHosts) {
```
lukecwik: Why don't we check that registerAllExtensions exists so that it fails at Dataflow definition time instead of later when the Dataflow is running.

sdk/src/main/java/com/google/cloud/dataflow/sdk/coders/Proto2Coder.java
```diff
@@ -0,0 +1,139 @@
...
+
+  /**
+   * Produces a new Proto2Coder instance, for a given Protobuf message class.
+   */
+  public static <T extends Message> Proto2Coder<T> of(Class<T> protoMessageClass) {
+    return new Proto2Coder<>(protoMessageClass);
+  }
+
+  private final String protoMessageClassName;
+  private final List<String> extensionClassNames = new ArrayList<String>();
```
lukecwik: Class is serializable, instead of storing the string name, just store the Class<?> and List<Class<?>>, saves you from having to look it up later with Class.forName

sdk/src/main/java/com/google/cloud/dataflow/sdk/coders/Proto2Coder.java
```diff
@@ -0,0 +1,139 @@
...
+import java.io.OutputStream;
+import java.lang.reflect.InvocationTargetException;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * An encoder using Google Protocol Buffers 2 binary format.
+ * <p>
+ * To learn more about Protocol Buffers, visit:
+ * https://developers.google.com/protocol-buffers
```
lukecwik: Please use an <a> tag for the link.

perezd:
All your comments have been addressed, PTAL, thanks!

lukecwik:
Thanks Derek for all the effort you put into making the Proto2Coder.

perezd:
Happy to help! Glad to see it made its way in.

PR #8, opened 2015-02-20
Make fn in Combine.PerKey Non-Transient, Add Getter

aljoscha:
We are implementing a PipelineRunner for Apache Flink: http://flink.apache.org. Making the KeyedCombineFn accessible in Combine.PerKey would allow us to translate it to a more efficient runtime operation.

### DIFF
--- a/sdk/src/main/java/com/google/cloud/dataflow/sdk/transforms/Combine.java
+++ b/sdk/src/main/java/com/google/cloud/dataflow/sdk/transforms/Combine.java
@@ -910,6 +910,13 @@ public class Combine {
       this.fn = fn;
     }
 
+    /**
+     * Returns the KeyedCombineFn used by this Combine operation.
+     */
+    public KeyedCombineFn<? super K, ? super VI, ?, VO> getFn() {
+      return fn;
+    }
+
     @Override
     public PCollection<KV<K, VO>> apply(PCollection<KV<K, VI>> input) {
       return input


davorbonaci:
Hi Aljoscha,
Thanks for this pull request.

There are some trade-offs involved in accepting this. Thus, before accepting this change, we'd like to make sure the benefits exceed the costs incurred and consider any possible alternatives.

Would you be able to share with us how and why this change would benefit your project? If possible, please include any relevant architectural details, e.g., when/where your translation to runtime operation occurs. Are you aware of any alternatives that would achieve the same or a similar benefit?

Thanks!

aljoscha:
Hi Davor,
the translation happens while traversing the Pipeline in our version of the DataflowPipelineTranslator. The problem is that Combine.PerKey gets translated to a GroupByKey followed by a GroupedValues that does the aggregation of the values. In Flink, we have a special operation that does grouping and aggregation using a User Function in one step (GroupReduce) but we can only use it if we have the grouping and the combine function from the user in one place.

The alternative I see is to check in the translation of the GroupedValues whether the input Transform is a GroupByKey and then overwriting the result of the previous two-step operation with a new special GroupReduce operation. This could lead to duplicate work, however, since the result of the GroupByKey operation could be used by several consumers, each of those would now basically perform it's own grouping by key.

Regards,
Aljoscha

robertwb:
Making this visible to runner implementors, as implemented in this pull request, is the right way to support this kind of behavior.

davorbonaci:
If I may ask, what is the reasoning behind removal of the 'transient' keyword? Does some translation processing perhaps occur in the backend?

aljoscha:
I removed it just to make sure that the field stays there. All the translation is handled while traversing the Pipeline via traverseTopologically(). So yes, transient could be added again.

davorbonaci:
Could you please add the 'transient' keyword back? We'd prefer not to increase the size of backend requests unnecessarily.

After this small update, the pull request is approved for merge.

Additionally, for legal reasons, could you take care of the CLA mentioned in our [contributing](https://github.com/GoogleCloudPlatform/DataflowJavaSDK/blob/master/CONTRIBUTING.md) file. This is a mandatory one-time process for every contributor who is not a Google employee.

Thanks.

aljoscha:
Thanks, I updated this pull request and signed the CLA. Will you automatically get to see my CLA or do I have to do something for that? I gave my github account name in the CLA.

davorbonaci:
CLA confirmed. Pull request merged.

Thanks!

aljoscha:
Thanks for merging!

PR #10, opened 2015-03-03
Generic implementation of inner and outer left/right join.

mrunesson:
This is a patch to make it easier for user of DataFlow to do joins in the most common
circumstances. It implements inner join and outer left and right join.

In the cases of outer join. Since null cannot be serialized the user have to provide a
value that represent null for that particular usecase.

### DIFF
--- a/contrib/join-library/README.md
+++ b/contrib/join-library/README.md
@@ -1,14 +1,14 @@
 Join-library
 ============
 
-Join-library provide inner join, outer left and right join functions to
-Google DataFlow. The aim is to simplify the most common cases of join to a
+Join-library provides inner join, outer left and right join functions to
+Google Cloud Dataflow. The aim is to simplify the most common cases of join to a
 simple function call.
 
 The functions are generic so it supports join of any types supported by
-DataFlow. Input to the join functions are PCollections of Key/Values. Both the
+Dataflow. Input to the join functions are PCollections of Key/Values. Both the
 left and right PCollections need the same type for the key. All the join
-functions returns a Key/Value where Key is the join key and value is
+functions return a Key/Value where Key is the join key and value is
 a Key/Value where the key is the left value and right is the value.
 
 In the cases of outer join, since null cannot be serialized the user have
@@ -22,4 +22,4 @@ Example how to use join-library:
     PCollection<KV<String, KV<String, Long>>> joinedPcollection =
       Join.innerJoin(leftPcollection, rightPcollection);
 
-Questions or comments: <M.Runesson@gmail.com>
+Questions or comments: <M.Runesson [at] gmail [dot] com>
--- a/contrib/join-library/README.md
+++ b/contrib/join-library/README.md
@@ -1,0 +1,25 @@
+Join-library
+============
+
+Join-library provide inner join, outer left and right join functions to
+Google DataFlow. The aim is to simplify the most common cases of join to a
+simple function call.
+
+The functions are generic so it supports join of any types supported by
+DataFlow. Input to the join functions are PCollections of Key/Values. Both the
+left and right PCollections need the same type for the key. All the join
+functions returns a Key/Value where Key is the join key and value is
+a Key/Value where the key is the left value and right is the value.
+
+In the cases of outer join, since null cannot be serialized the user have
+to provide a value that represent null for that particular use case.
+
+Example how to use join-library:
+
+    PCollection<KV<String, String>> leftPcollection = ...
+    PCollection<KV<String, Long>> rightPcollection = ...
+
+    PCollection<KV<String, KV<String, Long>>> joinedPcollection =
+      Join.innerJoin(leftPcollection, rightPcollection);
+
+Questions or comments: <M.Runesson@gmail.com>
--- a/contrib/join-library/pom.xml
+++ b/contrib/join-library/pom.xml
@@ -6,9 +6,9 @@
   <modelVersion>4.0.0</modelVersion>
 
   <groupId>com.google.cloud.dataflow</groupId>
-  <artifactId>join-library</artifactId>
+  <artifactId>google-cloud-dataflow-java-contrib-join-library</artifactId>
   <name>Join library</name>
-  <description>Library with generic join functions for DataFlow.</description>
+  <description>Library with generic join functions for Dataflow.</description>
   <version>0.0.2-SNAPSHOT</version>
   <packaging>jar</packaging>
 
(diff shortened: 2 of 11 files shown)


davorbonaci:
Hi Magnus,
Thanks for this contribution and feedback. We appreciate it.

This pull request, however, requires some additional discussion before we can triage it. We'll get back to you shortly with our thoughts.

mrunesson:
You are welcome.

There are some things I have been thinking on that will make this code more usable and easy to use. First is support for a tuples. That gives the possibility to not have to use a KV to store the two values.
The second thing is make use of Optional to handle the cases of Outer joins and null.

davorbonaci:
We have discussed this internally and we'd like to propose to make this the first community-contributed module in our [`contrib`](https://github.com/GoogleCloudPlatform/DataflowJavaSDK/tree/master/contrib) directory.

Improved Join support is certainly something that is on our radar, but we aren't ready to commit to a particular API just yet. Including it now and potentially changing it later may cause user frustration. Thus, we think starting from a separate module is the right approach at this moment. As soon as our thinking on Join support (and the module itself) matures, we should reconsider pulling it back into the mainline SDK.

If this is something you are comfortable with, here are the steps:
* Create a directory under `contrib`. Call it, perhaps, `join-library`.
* In the root of the directory, add `pom.xml`. You may or may not choose to reference our parent [`pom.xml`](https://github.com/GoogleCloudPlatform/DataflowJavaSDK/blob/master/pom.xml) and [`checkstyle.xml`](https://github.com/GoogleCloudPlatform/DataflowJavaSDK/blob/master/checkstyle.xml).
* In the root of the directory, add `README.md` with as much usage information as possible.
* Move `Join.java` and `*JoinTest.java` into that directory. Please organize it in some reasonable way. For example, add sub-directories `src/main/java/com/...` and `src/test/java/com/...`.

Regarding changes to `.gitignore`, we'd be happy to merge them right away. Could you please move them a separate pull request?

mrunesson:
Sounds totally reasonable. I'll move the join to contrib and do the gitignore to another pullrequest.

mrunesson:
I have moved the join to contrib/join-library after a rebase from master and must messed up something. Let me know if this works or I should do a clean pull request.

mrunesson:
Join code is moved to contrib/join-library and README and pom added. Gitignore removed from this pullrequest.. 

Let me know if you have any comments.

davorbonaci:
Thanks. CLA confirmed. Review in progress.

contrib/join-library/README.md
```diff
@@ -0,0 +1,25 @@
+Join-library
+============
+
+Join-library provide inner join, outer left and right join functions to
```
davorbonaci: provide -> provides

contrib/join-library/README.md
```diff
@@ -0,0 +1,25 @@
...
+Google DataFlow. The aim is to simplify the most common cases of join to a
```
davorbonaci: Google DataFlow --> Google Cloud Dataflow

contrib/join-library/README.md
```diff
@@ -0,0 +1,25 @@
...
+simple function call.
+
+The functions are generic so it supports join of any types supported by
+DataFlow. Input to the join functions are PCollections of Key/Values. Both the
```
davorbonaci: DataFlow -> Dataflow

contrib/join-library/README.md
```diff
@@ -0,0 +1,25 @@
...
+left and right PCollections need the same type for the key. All the join
+functions returns a Key/Value where Key is the join key and value is
```
davorbonaci: returns -> return

contrib/join-library/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+Example how to use join-library:
+
+    PCollection<KV<String, String>> leftPcollection = ...
+    PCollection<KV<String, Long>> rightPcollection = ...
+
+    PCollection<KV<String, KV<String, Long>>> joinedPcollection =
+      Join.innerJoin(leftPcollection, rightPcollection);
+
+Questions or comments: <M.Runesson@gmail.com>
```
davorbonaci: You may want to enter your email address like this: M.Runesson [at] gmail [dot] com, to reduce the possibility of your email address being picked by automatic crawlers.

contrib/join-library/pom.xml
```diff
@@ -0,0 +1,91 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>com.google.cloud.dataflow</groupId>
+  <artifactId>join-library</artifactId>
+  <name>Join library</name>
+  <description>Library with generic join functions for DataFlow.</description>
```
davorbonaci: DataFlow -> Dataflow.

contrib/join-library/pom.xml
```diff
@@ -0,0 +1,91 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>com.google.cloud.dataflow</groupId>
+  <artifactId>join-library</artifactId>
```
davorbonaci: I'd put artifactId as `google-cloud-dataflow-java-contrib-join-library` to be inline with other artifacts.

contrib/join-library/src/main/java/com/google/cloud/dataflow/contrib/joinlibrary/Join.java
```diff
@@ -0,0 +1,186 @@
+/*
+ * Copyright (C) 2014 Google Inc.
```
davorbonaci: I'd suggest changing this to: `Copyright (C) 2015 The Google Cloud Dataflow Authors`, here and in other `*.java` files, to acknowledge your contributions. If checkstyle complains, feel free to fix the regex (or I can do it).

mrunesson:
Thanks for the review comments. I have adjusted to them. 

This requires the change in https://github.com/GoogleCloudPlatform/DataflowJavaSDK/pull/13 to build.

davorbonaci:
This pull request is now fully merged, with a few small tweaks.

Thanks Magnus for contributing our first community-sourced module! Much appreciated. This will be very useful to many of our users. Naturally, any future improvements, bug fixes, etc. are always welcome.

PR #11, opened 2015-03-06
Update Proto2Coder.java

earhart:
Add serialVersionUID for this serializable class.


davorbonaci:
CLA not needed. Merged. Thanks Rob.

PR #21, opened 2015-05-08
Fixed error in examples of the Filter JavaDoc.

alexvanboxel:
I corrected an error in the filter JavaDoc. The example showed comparing numbers but the type was string. This generated confusion so I changed it to Integers.


davorbonaci:
Hi Alex,
Indeed, the example in Javadoc is confusing. Thanks for fixing it.

Perhaps it is obvious, but let me quickly note that `Filter.lessThan`, `Filter.greaterThan`, etc. support filtering of any Comparable type, including String and Integer. The issue is in the Javadoc example only.

CLA confirmed. We will merge this pull request shortly.

alexvanboxel:
Yes the filtering on Comparable type was obvious enough. The error in the example didn't compute in my head :-).

davorbonaci:
Merged. Thanks!

PR #46, opened 2015-07-16
Check if PubSUb subscription is null, before checking if it's valid

i-maravic:
This targets only execution via DirectPipelineRunner class.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 target/
+sdk/dependency-reduced-pom.xml
 
 # Ignore IntelliJ files.
 .idea/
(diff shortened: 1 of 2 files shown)


i-maravic:
I signed it!

Spotify is company's name.

davorbonaci:
Hi Igor,
Sorry for the delay in response on this pull request.

It seems like we have already fixed this bug in the existing commit cf946ca174efe80a02e8815356f8a19b0f5b203e. Sorry about that.

Hopefully, this issue should now be resolved. Please let us know if that's not the case.

Separately, we'd like to take the change to `.gitignore`. Could you please send a separate pull request for that change?

Thanks!

PR #61, opened 2015-09-29
Contrib module for natsio

tyagihas:


tyagihas:
I signed it!

contrib/natsio/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+Install and launch [NATS server](http://nats.io/docs/#gnatsd), typically on Google Compute Engine.
+```bash
+% gnatsd
+```
+
+Run tests with Maven. The following command will launch a producer and consumer job respectively.<BR>
+The producer publishes messages and the consumer will receive them via NATS server.
+```bash
+<<<<<<< HEAD
```
dhalperi: merge conflicts?

contrib/natsio/pom.xml
```diff
@@ -0,0 +1,58 @@
...
+				</executions>
+			</plugin>
+		</plugins>
+	</build>
+
+  <dependencies>
+  	<dependency>
+  		<groupId>com.github.tyagihas</groupId>
+  		<artifactId>java_nats</artifactId>
+  		<version>0.5.1</version>
```
dhalperi: Is this needed?

contrib/natsio/src/main/java/com/google/cloud/dataflow/contrib/natsio/NatsIO.java
```diff
@@ -0,0 +1,359 @@
...
+import java.util.concurrent.ConcurrentLinkedQueue;
+
+import org.joda.time.Duration;
+import org.joda.time.Instant;
+import org.nats.Connection;
+import org.nats.MsgHandler;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.cloud.dataflow.contrib.natsio.NatsIO.Read.NatsSource.CommitMark;
```
dhalperi: I'm confused why you have to `import` something declared in this file. Is this for brevity later?

contrib/natsio/pom.xml
```diff
@@ -0,0 +1,58 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
```
dhalperi: It would be good to follow the model set out in the [contrib join library](https://github.com/GoogleCloudPlatform/DataflowJavaSDK/blob/master/contrib/join-library/pom.xml). Specifically

- Artifact naming
- License information
- Compiler source and target versions (1.7, probably)
- optionally, Checkstyle

- review dependencies

contrib/natsio/pom.xml
```diff
@@ -0,0 +1,58 @@
...
+  	</dependency>
+  	<dependency>
+  		<groupId>com.google.cloud.dataflow</groupId>
+  		<artifactId>google-cloud-dataflow-java-sdk-all</artifactId>
+  		<version>[1.0.0,2.0.0)</version>
+  	</dependency>
+  	<dependency>
+  		<groupId>org.apache.maven.plugins</groupId>
+  		<artifactId>maven-surefire-plugin</artifactId>
+  		<version>2.18.1</version>
```
dhalperi: is this needed here? I think this just needs to appear where it is above.

contrib/natsio/src/main/java/com/google/cloud/dataflow/contrib/natsio/NatsIO.java
```diff
@@ -0,0 +1,359 @@
...
+			private Connection conn = null;
+			private String subject;
+			
+			public NatsSource(String subject, Properties props) {
+				this.subject = subject;
+				this.props = props;
+			}
+			
+			@Override
+			public NatsReader createReader(PipelineOptions arg0, UnboundedSource.CheckpointMark arg1) {
```
dhalperi: in general, would prefer better variable names throughout than `arg0`, `arg1`. Any reason not to use the original names from the overridden function?

contrib/natsio/README.md
tyagihas: Fixed.

contrib/natsio/pom.xml
tyagihas: Yes. It's a client module for NATS server.

contrib/natsio/pom.xml
tyagihas: Removed the entry.

contrib/natsio/src/main/java/com/google/cloud/dataflow/contrib/natsio/NatsIO.java
tyagihas: Fixed.

contrib/natsio/src/main/java/com/google/cloud/dataflow/contrib/natsio/NatsIO.java
```diff
@@ -0,0 +1,358 @@
...
+
+				@Override
+				public boolean start() throws IOException {
+					if (conn == null) {
+						try {
+							conn = Connection.connect(props);
+						} catch (InterruptedException e) {
+							e.printStackTrace();
+						}
+					}
```
dhalperi: It looks like a `NullPointerException` can happen here when the thread is interrupted. This seems like the wrong handling of the `InterruptedException`?

tyagihas: Agree. I will change to throw an IOException.

dhalperi: Since this is for a source/sink, what you want to do is effectively tell the caller above that you were closed by an interrupt. 

Throwing something like [`ClosedByInterruptException`](http://docs.oracle.com/javase/7/docs/api/java/nio/channels/ClosedByInterruptException.html) seems to make sense after cleaning up any resources.

We will fail the task, and restart the write operation again (on this or another worker).

tyagihas: Yes, that makes more sense. I will change it to throw ClosedByInterruptException.

dhalperi:
Hi @tyagihas:

I am encountering the following error installing this module:

```
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /home/dhalperi/tyagihas/DataflowJavaSDK/contrib/natsio/src/main/java/com/google/cloud/dataflow/contrib/nats
io/example/NatsIOBench.java:[14,48] cannot find symbol
  symbol:   class NatsIOTest
  location: package com.google.cloud.dataflow.contrib.natsio
[ERROR] /home/dhalperi/tyagihas/DataflowJavaSDK/contrib/natsio/src/main/java/com/google/cloud/dataflow/contrib/nats
io/example/NatsIOBench.java:[26,67] cannot find symbol
  symbol:   class NatsIOTest
  location: class com.google.cloud.dataflow.contrib.natsio.example.NatsIOBench
[INFO] 2 errors 
[INFO] -------------------------------------------------------------
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 5.870s
[INFO] Finished at: Mon Oct 19 21:13:44 UTC 2015
[INFO] Final Memory: 19M/56M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.2:compile (default-compile) on proj
ect google-cloud-dataflow-java-contrib-natsio: Compilation failure: Compilation failure:
[ERROR] /home/dhalperi/tyagihas/DataflowJavaSDK/contrib/natsio/src/main/java/com/google/cloud/dataflow/contrib/nats
io/example/NatsIOBench.java:[14,48] cannot find symbol
[ERROR] symbol:   class NatsIOTest
[ERROR] location: package com.google.cloud.dataflow.contrib.natsio
[ERROR] /home/dhalperi/tyagihas/DataflowJavaSDK/contrib/natsio/src/main/java/com/google/cloud/dataflow/contrib/nats
io/example/NatsIOBench.java:[26,67] cannot find symbol
[ERROR] symbol:   class NatsIOTest
[ERROR] location: class com.google.cloud.dataflow.contrib.natsio.example.NatsIOBench
[ERROR] -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoFailureException
```

Here's a log of the important commands I have run, on a fresh Ubuntu 15.04 VM:

- sudo apt-get update && sudo apt-get install -y git maven openjdk-8-jdk
- git clone https://github.com/GoogleCloudPlatform/DataflowJavaSDK.git && cd DataflowJavaSDK && mvn install && cd ~
- git clone https://github.com/tyagihas/DataflowJavaSDK.git natsio && cd natsio/contrib/natsio && mvn install

I suspect you want the NatsIOBench class to be in the `test` rather than in `main`. Normally, code in `main` should not depend on code in `test`.

dhalperi:
Assigning back to you for now, please reassign to me once fixed :).

dhalperi:
Hmm, can't re-assign to you, so please just bump the request once fixed. Thanks.

tyagihas:
Fixed the class reference. NatsIOBench shouldn't include reference to NatsIOTest. 
I also added some default test properties for maven test and only three parameters are mandatory for minimum testing.

mvn install -Dproject="project id" -DstagingLocation="bucket" -Dnats.servers=nats://"GCE private ip":4222

dhalperi:
These are the errors when I run `mvn install` with no args. I see above that I need to set some values, but the error messages could be significantly more helpful. (optional suggestion!)

```
Tests run: 2, Failures: 0, Errors: 2, Skipped: 0, Time elapsed: 0.023 sec <<< FAILURE! - in com.google.cloud.dataflow.cont
rib.natsio.NatsIOTest
publishSubscribe(com.google.cloud.dataflow.contrib.natsio.NatsIOTest)  Time elapsed: 0.002 sec  <<< ERROR!
java.lang.NullPointerException: null
        at java.util.Hashtable.put(Hashtable.java:459)
        at java.util.Properties.setProperty(Properties.java:166)
        at com.google.cloud.dataflow.contrib.natsio.NatsIOTest.<init>(NatsIOTest.java:65)
simplePublish(com.google.cloud.dataflow.contrib.natsio.NatsIOTest)  Time elapsed: 0.001 sec  <<< ERROR!
java.lang.NullPointerException: null
        at java.util.Hashtable.put(Hashtable.java:459)
        at java.util.Properties.setProperty(Properties.java:166)
        at com.google.cloud.dataflow.contrib.natsio.NatsIOTest.<init>(NatsIOTest.java:65)
```

dhalperi:
I am now able to get the test to run. I had to pull some tricks to get the gcloud credentials to work locally.

As of release 1.2.0, including Dataflow does not include an slf4j logger. So, when I run the test, I see this:

```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

which means that no logs will be shown. It would be good to add a default logger to the `pom.xml`, e.g.,

```xml
<dependency>
	<groupId>org.slf4j</groupId>
	<artifactId>slf4j-jdk14</artifactId>
	<version>1.7.7</version>
</dependency>
```

to get the old behavior that uses the default JDK logger.

dhalperi:
I ran the job as follows: `mvn -Dproject="$PROJECT" -DstagingLocation="$STAGING" -D"nats.servers=nats://0.0.0.0:4222" test`

Two Dataflow jobs start in my project, but they never make any progress and have not stopped after at least 10 minutes of running. In Cloud Logging, I see some very confusing logs, attached below.

It looks like some of the code needs to use a `org.slf4h.Logger` instead of `System.err/out` to report errors. Maybe this is coming from `org.nats.Connection` in this stack trace? https://github.com/tyagihas/java_nats/blob/master/src/main/java/org/nats/Connection.java#L215

This failure is somehow not failing the job, and I think it should.

dhalperi:
Both jobs did fail after 15 minutes, I guess because it was retried 4 times.

contrib/natsio/README.md
```diff
@@ -0,0 +1,23 @@
...
+```
+
+Run tests with Maven. The following command will launch a producer and consumer job respectively.<BR>
+The producer publishes messages and the consumer will receive them via NATS server.
+
+"stagingLocation", "project" and "nats.servers" are mandatory properties to run a test.
+```bash
+% mvn test -Dtest=NatsIOTest#publishSubscribe -DstagingLocation=gs://<bucket> ¥
+-Dproject=<project id> -Dnats.servers=nats://<server>:4222 -Dnats.queue=queue1 ¥
+-Dloop=30000 -Dinterval=0 -Dsubjects=test1 -Dconsumers=1 -Dproducers=1 -Dnats.maxRecords=20000 ¥
```
dhalperi: This symbol that is rendering as a Yen sign for me -- is this a Japanese keyboard version of `\` to mean "continues on next line"?

contrib/natsio/src/test/java/com/google/cloud/dataflow/contrib/natsio/NatsIOTest.java
```diff
@@ -0,0 +1,145 @@
...
+			.apply(ParDo.of(new DoFn<String, String>() {
+				public void processElement(DoFn<String, String>.ProcessContext c) throws Exception {
+					try {
+						Connection conn = Connection.connect(props);
+						for(int i = 0; i < loop; i++) {
+							conn.publish(subject, Integer.toString(i));
+							Thread.sleep(interval);
+						}
+					} catch (IOException e) {
+						e.printStackTrace();
```
dhalperi: It looks like the desired behavior here is to ignore errors, right? (You print, but don't throw). In that case, it should be something like:

```java
LOG.error("Ignoring exception in processElement", e);
```

In general, this does not seem very informative -- it can go wrong and we don't even learn how many elements were output.

contrib/natsio/src/test/java/com/google/cloud/dataflow/contrib/natsio/NatsIOTest.java
```diff
@@ -0,0 +1,145 @@
...
+					} catch (InterruptedException e) {
+						e.printStackTrace();
```
dhalperi: Should probably throw here.

contrib/natsio/src/test/java/com/google/cloud/dataflow/contrib/natsio/NatsIOTest.java
```diff
@@ -0,0 +1,145 @@
...
+		BlockingDataflowPipelineOptions producerOptions = buildOptions();
+		producerOptions.setJobName("producer-" + System.currentTimeMillis());
+		
+		Pipeline p2 = Pipeline.create(producerOptions);
+		p2
+			.apply(Create.of(subject)).setCoder(StringUtf8Coder.of())
+			.apply(ParDo.of(new DoFn<String, String>() {
+				public void processElement(DoFn<String, String>.ProcessContext c) throws Exception {
+					try {
+						Connection conn = Connection.connect(props);
```
dhalperi: do we need to close this connection?

contrib/natsio/src/test/java/com/google/cloud/dataflow/contrib/natsio/NatsIOTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+		// Starting a producer
+		BlockingDataflowPipelineOptions producerOptions = buildOptions();
+		producerOptions.setJobName("producer-" + System.currentTimeMillis());
+		
+		Pipeline p2 = Pipeline.create(producerOptions);
+		p2
+			.apply(Create.of(subject)).setCoder(StringUtf8Coder.of())
+			.apply(ParDo.of(new DoFn<String, String>() {
+				public void processElement(DoFn<String, String>.ProcessContext c) throws Exception {
```
dhalperi: It would be good to actually make this a static class with a useful name, to explain what this `DoFn` does.

Then you can move `conn = Connection.connect(props)` to `startBundle` and can close `conn` in `finishBundle`, and that better matches Dataflow SDK style.

contrib/natsio/src/test/java/com/google/cloud/dataflow/contrib/natsio/NatsIOTest.java
```diff
@@ -0,0 +1,145 @@
...
+					try {
+						Connection conn = Connection.connect(props);
+						for(int i = 0; i < loop; i++) {
+							conn.publish(subject, Integer.toString(i));
```
dhalperi: instead of `subject`, here, shouldn't you use `c.element()`?

contrib/natsio/src/test/java/com/google/cloud/dataflow/contrib/natsio/NatsIOTest.java
```diff
@@ -0,0 +1,145 @@
...
+							Thread.sleep(interval);
+						}
+					} catch (IOException e) {
+						e.printStackTrace();
+					} catch (InterruptedException e) {
+						e.printStackTrace();
+					}		
+					c.output("done");
```
dhalperi: In summary, would like this message to list the number of messages (successfully!) sent, and the `subject` name to which they were sent.

contrib/natsio/src/test/java/com/google/cloud/dataflow/contrib/natsio/NatsIOTest.java
```diff
@@ -0,0 +1,145 @@
...
+				public void processElement(DoFn<String, KV<String, String>>.ProcessContext c) throws Exception {
+					c.output(KV.of(c.element(), "hello world!!!"));
+				}
+			}))
+			.apply(NatsIO.Write.to(props));
+		p.run();
+	}
+	
+	@Test
+	public void publishSubscribe() throws InterruptedException {
```
dhalperi: I don't understand why these are `@Test` unit tests instead of standalone executables we run. Compare to `DataflowWordCount` --  this is in an `examples` folder.

Unit tests, by definition, should have no external dependencies.

Similarly, I should be able to run `mvn install` with no arguments and no network, and this should succeed (as long as I've downloaded all the .jars the project depends on).

contrib/natsio/src/main/java/com/google/cloud/dataflow/contrib/natsio/NatsIO.java
```diff
@@ -0,0 +1,359 @@
...
+import com.google.common.base.Preconditions;
+
+/**
+ * <p>{@link NatsIO} provides an API to publish messages to and subscribe subjects on <A href="http://nats.io">NATS server</A>.
+ * 
+ * <p>NATS server is a high-performance and lightweight cloud native messaging system. A message is a simple string and can be exchanged between
+ * clients written in various programming languages.
+ * 
+ * <p>NatsIO API requires <A href="http://nats.io">NATS server</A> running and accessible from Dataflow workers prior to job submissions. Typically,
+ * NATS server is expected to run on <A href="https://cloud.google.com/compute/">Google Compute Engine</A>.
```
dhalperi: Is this a hard requirement, or can NATS server run anywhere? What makes it need to be on GCE specifically?

contrib/natsio/src/main/java/com/google/cloud/dataflow/contrib/natsio/NatsIO.java
```diff
@@ -0,0 +1,359 @@
...
+		}		
+		
+		public static class NatsSource extends UnboundedSource<KV<String,String>, UnboundedSource.CheckpointMark> {
+			private static final long serialVersionUID = 1L;
+			
+			private Properties props = null;
+			private Connection conn = null;
+			private String subject;
+			
+			public NatsSource(String subject, Properties props) {
```
dhalperi: check that neither `subject` nor `props` is `null` here?

contrib/natsio/src/main/java/com/google/cloud/dataflow/contrib/natsio/NatsIO.java
```diff
@@ -0,0 +1,359 @@
...
+			public NatsReader createReader(PipelineOptions options, UnboundedSource.CheckpointMark mark) {
+				return new NatsReader(subject);
+			}
+
+			@Override
+			public List<? extends UnboundedSource<KV<String,String>, UnboundedSource.CheckpointMark>> generateInitialSplits(
+					int numSplits, PipelineOptions options) throws Exception {
+				List<NatsSource> results = new ArrayList<NatsSource>();
+				// Create a NatsSource instance for each split.
+				int split = numSplits / 3;
```
dhalperi: why divided by 3?

What happens if `numSplits` is `1` or `2`?

contrib/natsio/src/main/java/com/google/cloud/dataflow/contrib/natsio/NatsIO.java
```diff
@@ -0,0 +1,359 @@
...
+				for(int i = 0; i < split; i++)
+					results.add(new NatsSource(subject, props));
```
dhalperi: How does this implement splitting?

contrib/natsio/src/main/java/com/google/cloud/dataflow/contrib/natsio/NatsIO.java
```diff
@@ -0,0 +1,359 @@
...
+			}
+				
+			private class NatsReader extends UnboundedReader<KV<String,String>> {
+				private String subject;
+				private Integer sid;
+				private ConcurrentLinkedQueue<KV<String,String>> queue;
+				private KV<String,String> current;
+				
+				public NatsReader(String subject) {
+					this.subject = subject;
```
dhalperi: verify that `subject` is not null?

contrib/natsio/src/main/java/com/google/cloud/dataflow/contrib/natsio/NatsIO.java
```diff
@@ -0,0 +1,359 @@
...
+			public static class CommitMark implements UnboundedSource.CheckpointMark, Serializable {
+				private static final long serialVersionUID = 2L;
+
+				public void finalizeCheckpoint() throws IOException {}		
+			}
+				
+			private class NatsReader extends UnboundedReader<KV<String,String>> {
+				private String subject;
+				private Integer sid;
+				private ConcurrentLinkedQueue<KV<String,String>> queue;
```
dhalperi: make `subject` and `queue` final?

contrib/natsio/src/main/java/com/google/cloud/dataflow/contrib/natsio/NatsIO.java
```diff
@@ -0,0 +1,359 @@
...
+				public NatsReader(String subject) {
+					this.subject = subject;
+					queue = new ConcurrentLinkedQueue<KV<String,String>>();
+				}
+				
+				@Override
+				public boolean advance() throws IOException {
+					// Polling a queue to check whether an update is available in the queue.
+					current = queue.poll();
+					if (current == null)
```
dhalperi: totally optional, but Google java style is to always include `{}` with `if` statements or `for` loops, etc. Just helps prevent errors.

contrib/natsio/README.md
tyagihas: Yes, it's same meaning as backslash.

contrib/natsio/src/test/java/com/google/cloud/dataflow/contrib/natsio/NatsIOTest.java
tyagihas: I can maintain a counter and log a number of processed and failed records at the end.

contrib/natsio/src/test/java/com/google/cloud/dataflow/contrib/natsio/NatsIOTest.java
tyagihas: Will change it.

contrib/natsio/src/test/java/com/google/cloud/dataflow/contrib/natsio/NatsIOTest.java
tyagihas: Will add it.

contrib/natsio/src/test/java/com/google/cloud/dataflow/contrib/natsio/NatsIOTest.java
tyagihas: Agree. It looks much better.

contrib/natsio/src/test/java/com/google/cloud/dataflow/contrib/natsio/NatsIOTest.java
tyagihas: Yes, fixed.

contrib/natsio/src/test/java/com/google/cloud/dataflow/contrib/natsio/NatsIOTest.java
tyagihas: Yes, sounds good.

contrib/natsio/src/test/java/com/google/cloud/dataflow/contrib/natsio/NatsIOTest.java
tyagihas: Yes, I agree. I will make it as a standalone application.

contrib/natsio/src/main/java/com/google/cloud/dataflow/contrib/natsio/NatsIO.java
tyagihas: This isn't a hard requirement. As long as it's reachable from GCE, it can run anywhere. Ideally, it should run closer location for lower latency.

dhalperi:
GitHub notifications don't come when patches are pushed; please comment once code is updated next.

Thanks!

contrib/natsio/src/main/java/com/google/cloud/dataflow/contrib/natsio/NatsIO.java
tyagihas: Will check and throw an exception.

contrib/natsio/src/main/java/com/google/cloud/dataflow/contrib/natsio/NatsIO.java
tyagihas: I can't remember exactly in which source file, but numSplits is always passed as equal to 3 times of a number of workers.
Has this been changed in the latest build?

contrib/natsio/src/main/java/com/google/cloud/dataflow/contrib/natsio/NatsIO.java
tyagihas: You can make use of this behavior when a queue group is specified for load balancing. Multiple partitions subscribe to a single subject and whenever an update occurs, the server pushes an update to only one partition.

contrib/natsio/src/main/java/com/google/cloud/dataflow/contrib/natsio/NatsIO.java
tyagihas: Will check and throw a NullPointerException.

contrib/natsio/src/main/java/com/google/cloud/dataflow/contrib/natsio/NatsIO.java
tyagihas: Will change it.

contrib/natsio/src/main/java/com/google/cloud/dataflow/contrib/natsio/NatsIO.java
tyagihas: Thanks for the tip. Will conform with Google style then.

tyagihas:
Thank you for all the feedbacks. I've just pushed the latest code.

contrib/natsio/src/main/java/com/google/cloud/dataflow/contrib/natsio/NatsIO.java
dhalperi: I believe you're looking at [this code](https://github.com/GoogleCloudPlatform/DataflowJavaSDK/blob/198a00ebde74ec55779b997f1e961e66101c4093/sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/dataflow/CustomSources.java#L351) for the Dataflow runner.

Developers should only rely on the publicly documented interfaces, e.g., in this case, the Javadoc on `UnboundedSource.generateInitialSplits`.

Code inside the `runners` or `util` packages is not part of the "public API", and is generally subject to change at any time.  Even if that code has some current behavior, you cannot assume it will hold in the future. We don't want you to have to rewrite code if internal implementation details change!

The other reason to stick to the publicly documented API is that there are _other Dataflow runners_, e.g., Cloudera's Spark-based runner. They may not provide the same implementation we do internally.

(Finally, that code has 3 different possible return paths, and in only 1 of them is the result guaranteed to be both larger than 2 and a multiple of 3. The user may set `--maxNumWorkers=1`, for example, which would lead to a `numSplits` of `1` and thus you'd compute `0` here!)

tyagihas: Understood. The code (div 3) is removed.

tyagihas:
Thank you for the comments. I just push a new commit.

tyagihas:
Hi @dhalperi 

It's been a while since last update. Did you have time to look at the latest commit?

contrib/natsio/src/main/java/com/google/cloud/dataflow/contrib/natsio/NatsIO.java
dhalperi: please update the comment to remove this requirement -- should work with any NATS server.

Additionally, the job may not be running in GCE in unless the jobs is running in Google Cloud Dataflow – there are other runners.

contrib/natsio/src/main/java/com/google/cloud/dataflow/contrib/natsio/NatsIO.java
```diff
@@ -0,0 +1,395 @@
...
+		public static class NatsSource extends UnboundedSource<KV<String,String>, UnboundedSource.CheckpointMark> {
+			private static final long serialVersionUID = 1L;
+			
+			private Properties props = null;
+			private Connection conn = null;
+			private String subject;
+			
+			public NatsSource(String subject, Properties props) {
+				if (subject == null) {
+					throw new NullPointerException("subject");
```
dhalperi: would be simpler if you use Guava Preconditions:

```java
import static com.google.common.base.Preconditions.checkNotNull;

....

checkNotNull(subject, "subject");
...
```

contrib/natsio/src/main/java/com/google/cloud/dataflow/contrib/natsio/NatsIO.java
```diff
@@ -0,0 +1,395 @@
...
+				}
+				if (props == null) {
+					throw new NullPointerException("props");
+				}
+				if (props.getProperty("servers") == null) {
+					throw new NullPointerException("servers");
```
dhalperi: I would probably change this one to a `Preconditions.checkArgument` instead, since it's not a direct parameters.

```
import static com.google.common.base.Preconditions.checkArgument;

...

checkArgument(props.getProperty("servers") != null, "Expected property \"servers\" in the provided properties");
```

contrib/natsio/src/main/java/com/google/cloud/dataflow/contrib/natsio/NatsIO.java
```diff
@@ -0,0 +1,395 @@
...
+
+			@Override
+			public Coder<KV<String,String>> getDefaultOutputCoder() {
+				return KvCoder.of(NatsIO.DEFAULT_CODER, NatsIO.DEFAULT_CODER);
+			}
+
+			@Override
+			public void validate() {
+				Preconditions.checkNotNull(props.getProperty("servers"), "need to set the 'servers' property");
+				Preconditions.checkNotNull(subject, "need to set the 'subject'");
```
dhalperi: is there a way these can be `null` or missing given how the source is constructed?

contrib/natsio/src/main/java/com/google/cloud/dataflow/contrib/natsio/NatsIO.java
```diff
@@ -0,0 +1,395 @@
...
+				}
+
+				@Override
+				public KV<String,String> getCurrent() throws NoSuchElementException {
+					return current;
+				}
+
+				@Override
+				public Instant getCurrentTimestamp() throws NoSuchElementException {
+					return Instant.now();
```
dhalperi: This method does not meet its spec. Specifically, if the user calls calls this method twice without an intervening call to `advance()`, the source will return the same element (in `getCurrent`) -- and that element should have the same timestamp.

contrib/natsio/src/test/java/com/google/cloud/dataflow/contrib/natsio/NatsIOUnitTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+	@Test
+	public void testSinkNullServer() {
+		thrown.expect(NullPointerException.class);
+		thrown.expectMessage("servers");
+		Properties emptyProps = new Properties();
+		NatsSink sink = new NatsSink(emptyProps);
+		sink.validate(null);
+	}
+}
```
dhalperi: I still don't see any unit tests of the actual source.

There should be some way of setting up a fake/mock NATS.io server in a unit test and loading up some messages, and exercising the code that way.

- You should verify that it is possible to resume from checkpoints, so that the source is fault-tolerant.

- You should verify that even when split up, the source gets all messages exactly once.

- You should verify that when writing, the sink writes each element exactly once.

You might look at the [unit tests for `CountingSource`](https://github.com/GoogleCloudPlatform/DataflowJavaSDK/blob/master/sdk/src/test/java/com/google/cloud/dataflow/sdk/io/CountingSourceTest.java) to get some ideas.

dhalperi:
No movement for 60 days since last review, so I'm closing.

Note that we are moving active development to Apache Beam (incubating). If you wish to resubmit, please read the [Beam contribution guide](http://beam.incubator.apache.org/contribution-guide/).

PR #73, opened 2015-10-21
Fixed close not being called

elibixby:
`close()` is not called by wrapping `BoundedReadFromUnboundedSource` resulting in finalization behavior for the `UnboundedSource` not executing.


dhalperi:
Thanks Eli! We'll take care of testing in followup commit as needed.

PR #77, opened 2015-10-29
Travis: cache the Maven directory of downloaded jars

dhalperi:
Reduce failure rate of downloading from Maven

The first run should be slower, but subsequent runs should
be faster.


davorbonaci:
Let's not take this, given it would make the build less hermetic.

PR #82, opened 2015-11-11
worker: fix leak in StateSampler

danielnorberg:
Mismatching keys were being used for get and set in the statesByName map, causing the state/counter existence check to always fail and every invocation of stateForName() to create a new counter and state.

This manifests as a memory leak where the majority of the heap is consumed by the contents of statesByName and kindsByState, ultimately causing the worker to fail.


danielnorberg:
I signed it!

Spotify has signed a company wide CLA with Google.

davorbonaci:
CLA manually confirmed. Merged.

Thanks Daniel, much appreciated!

Issue #60, opened 2015-09-23
[Question] KafkaIO Sink/Source


kevinsookocheff-wf:
Hi, I've been working on a streaming KafkaIO.Read and KafkaIO.Write. From this conversation it seems like Google is already working on this and I don't want to duplicate effort. Can anyone comment on the status of this? I'd be happy to help get it out the door.

http://stackoverflow.com/questions/29893342/is-it-possible-to-use-kafka-with-google-cloud-dataflow

dhalperi:
@kevinsookocheff-wf: We are not working actively on Kafka connectors at this time. I would be happy to help review your code or answer questions are your implementation moves forward.

kevinsookocheff-wf:
I started working on this using the new consumer API available in Kafka 0.9. One question I had was which pattern to use to implement this. I could inherit from the abstract `Source` classes as described [here](https://cloud.google.com/dataflow/model/custom-io). I also see a lot of the standard IO features following a different pattern. The contrib modules currently in pull request mix both styles. Ultimately, to me it doesn't make much difference, I would just like to know what the recommended style is.

I am also interested in having both unbounded source and sinks and the abstract sink classes don't exist yet for the unbounded case.

dhalperi:
You are correct that our IO connectors do not have a consistent style: the guidelines for Custom IO you linked came later and are more refined. We do recommend that you follow the Custom IO guidelines for your sources and sinks.

We would like to provide a good abstraction for custom unbounded sinks, but you're right that this is not available right now.

kevinsookocheff-wf:
@dhalperi I've made a lot of progress on this but there are some parts of the custom source API that I'm not sure I understand with the current documentation. Particularly around failure scenarios.

What I've done so far is read messages from Kafka into a buffer to be processed with the Dataflow SDK. These messages have been acknowledged to Kafka as being read but I need a way to handle the case where I have acknowledged the messages but the Dataflow process fails and restarts before I'm able to process the messages. Is there a point in the UnboundedSource API that I can serialize unprocessed data and restore it when the service comes back up.

dhalperi:
Hi @kevinsookocheff-wf -- great to hear that you're making progress!

We expect that you will use checkpointing (see `getCheckpointMark` and other related functions) to include state you don't want to lose. The runner itself will call `getCheckpointMark` to ensure that data is persisted.

[...]

PR #96, opened 2016-01-20
Kafka protype.

kevinsookocheff-wf:
As requested by https://github.com/GoogleCloudPlatform/DataflowJavaSDK/issues/60#issuecomment-172500175

Contains a prototype implementation of reading from Kafka. Requires Kafka 0.9.0.0.

### DIFF
--- a/contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/coders/ConsumerRecordCoder.java
+++ b/contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/coders/ConsumerRecordCoder.java
@@ -1,0 +1,52 @@
+package com.google.cloud.dataflow.contrib.kafka.coders;
+
+import com.google.cloud.dataflow.sdk.coders.CustomCoder;
+import com.google.common.base.Charsets;
+import com.google.common.io.CharStreams;
+import com.google.gson.Gson;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.io.OutputStream;
+
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.codehaus.jackson.annotate.JsonCreator;
+
+/**
+ * A {@code ConsumerRecordCoder} encodes Kafka ConsumerRecord objects.
+ */
+public class ConsumerRecordCoder extends CustomCoder<ConsumerRecord<String, String>> {
+  private static final long serialVersionUID = 1L;
+
+
+  @JsonCreator
+  public static ConsumerRecordCoder of() {
+    return INSTANCE;
+  }
+
+  private static final ConsumerRecordCoder INSTANCE = new ConsumerRecordCoder();
+
+  private ConsumerRecordCoder() {}
+
+  @Override
+  public void encode(ConsumerRecord<String, String> value, OutputStream outStream, Context context)
+      throws IOException {
+    Gson gson = new Gson();
+    String json = gson.toJson(value);
+    outStream.write(json.getBytes(Charsets.UTF_8));
+  }
+
+  @Override
+  public ConsumerRecord<String, String> decode(InputStream inStream, Context context)
+      throws IOException {
+    String json;
+    try (InputStreamReader reader = new InputStreamReader(inStream, Charsets.UTF_8)){
+      json = CharStreams.toString(reader);
+    }
+    Gson gson = new Gson();
+    @SuppressWarnings("unchecked")
+    ConsumerRecord<String, String> record = gson.fromJson(json, ConsumerRecord.class);
+    return record;
+  }
+}
--- a/contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/io/KafkaCheckpoint.java
+++ b/contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/io/KafkaCheckpoint.java
@@ -1,0 +1,104 @@
+package com.google.cloud.dataflow.contrib.kafka.io;
+
+import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.LinkedBlockingQueue;
+
+import org.apache.kafka.clients.consumer.ConsumerRecord;
+import org.apache.kafka.clients.consumer.OffsetAndMetadata;
+import org.apache.kafka.common.TopicPartition;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.google.cloud.dataflow.sdk.io.UnboundedSource;
+import com.google.common.collect.Maps;
+
+/**
+ * A marker representing the progress and state of the {@link UnboundedReader}.
+ *
+ * Starts a separate thread to read messages from Kafka to populate a {@link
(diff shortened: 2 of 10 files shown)


kevinsookocheff-wf:
The CLA has been signed as a corporation (Workiva).

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/coders/ConsumerRecordCoder.java
```diff
@@ -0,0 +1,52 @@
...
+  private static final ConsumerRecordCoder INSTANCE = new ConsumerRecordCoder();
+
+  private ConsumerRecordCoder() {}
+
+  @Override
+  public void encode(ConsumerRecord<String, String> value, OutputStream outStream, Context context)
+      throws IOException {
+    Gson gson = new Gson();
+    String json = gson.toJson(value);
+    outStream.write(json.getBytes(Charsets.UTF_8));
```
dhalperi: might consider using versions that write to / read from streams directly ,e.g., https://sites.google.com/site/gson/streaming

(We actually use Jackson in Dataflow core SDK, not sure about pros/cons of Gson).

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/io/KafkaSource.java
```diff
@@ -0,0 +1,251 @@
...
+     *
+     * Kafka records do not have a natural timestamp so min value instead.
+     */
+    @Override
+    public Instant getCurrentTimestamp() throws NoSuchElementException {
+      ConsumerRecord<String, String> record = getCurrent();
+      if (record == null) {
+        throw new NoSuchElementException();
+      }
+      return BoundedWindow.TIMESTAMP_MIN_VALUE;
```
dhalperi: If you always return the `TIMESTAMP_MIN_VALUE` then I think your streaming pipeline will stall in practice.

A usual strategy is to let users specify a `SerializableFunction<T, Instant>` to pull the timestamp out of the element if there is one in the payload, or to use `Instant.now()` when the element is pulled from Kafka. (Note that successive calls to `getCurrentTimestamp()` should return the same value, so using `Instant.now()` here would not be semantically correct.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/io/KafkaCheckpoint.java
```diff
@@ -0,0 +1,104 @@
...
+
+  /**
+   * Return the current record.
+   */
+  public ConsumerRecord<String, String> current() throws NoSuchElementException {
+    if (current == null) {
+      throw new NoSuchElementException();
+    }
+
+    LOG.info("Current Record: topic={}, partition={}, offset={}, value={}",
```
dhalperi: drop to `debug` or lower? wrap in `if (LOG.isInfoEnabled()){}`? This will destroy pipeline perf otherwise.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/io/KafkaCheckpoint.java
```diff
@@ -0,0 +1,104 @@
...
+    return current;
+  }
+
+  /**
+   * Shutdown the producer thread.
+   */
+  public void shutdown() {
+    producer.shutdown();
+  }
+}
```
dhalperi: this is a really interesting way to design the checkpoint. Are there any guarantees that when the checkpoint is being serialized it will be a "consistent" checkpoint, or will Kafka be able to insert records to the different fields willy-nilly?

dhalperi: (aka, is the `finalizeCheckpoint()` function atomic?)

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/io/KafkaProducer.java
```diff
@@ -0,0 +1,212 @@
...
+      sessionTimeout = DEFAULT_SESSION_TIMEOUT;
+    }
+
+    Properties props = new Properties();
+    props.put("bootstrap.servers", bootstrapServers);
+    props.put("group.id", groupId);
+    props.put("session.timeout.ms", Integer.toString(sessionTimeout));
+    props.put("enable.auto.commit", "false");
+    props.put("key.deserializer", "org.apache.kafka.common.serialization.StringDeserializer");
+    props.put("value.deserializer", "org.apache.kafka.common.serialization.StringDeserializer");
```
dhalperi: might make sense to let the user pass in a `Properties` to override with.

contrib/kafka/src/test/java/com/google/cloud/dataflow/contrib/kafka/coders/ConsumerRecordCoderTest.java
```diff
@@ -0,0 +1,65 @@
...
+import org.junit.runners.JUnit4;
+
+/**
+ * Test case for {@link ConsumerRecordCoder}.
+ */
+@RunWith(JUnit4.class)
+public class ConsumerRecordCoderTest {
+
+  @Test
+  public void testDecodeEncodeEqual() throws Exception {
```
dhalperi: you're welcome to use functions from `CoderProperties` that implement a lot of these methods already.

Nice tests, though!

dhalperi:
Hi Kevin,

Thanks for this pull request. Some nice stuff in here and others will have a great foundation to build on.

I know it's a WIP and that you've moved on. I left some comments, but how you want to respond is up to you. You could make code changes, you could make TODOs, you could leave as-is untouched.

The level of polish on a contrib package is generally up to the author. Let me know when it's ready for either more review or merging.

Thanks,
Dan

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/io/KafkaCheckpoint.java
kevinsookocheff-wf: To be honest I'm not 100% happy with this design. There is probably a better way to manage the concurrent access to the Kafka thread I just haven't found it yet. 

The limitation with this design is that you may get duplicate records if a crash happens before committing. Otherwise it seems to be safe.

contrib/kafka/src/test/java/com/google/cloud/dataflow/contrib/kafka/coders/ConsumerRecordCoderTest.java
kevinsookocheff-wf: I had meant to leave a comment or PR about this. The utilities in the repo only work if your element overrides equals/hashcode. I also cannot access the standalone encode/decode utilities because they are private to the package.

kevinsookocheff-wf:
@dhalperi I will probably leave this as is (for now). If you prefer I can put this up on my personal Github rather than here.

dhalperi:
Thanks @kevinsookocheff-wf! I think we'll leave this here and open as a place for others who come looking.

dhalperi:
Hi Kevin,

KafkaIO is now in active development under PR #121 for BEAM-52. Would you be okay with closing this PR?

PR #103, opened 2016-01-23
fixes and enchancements for contrib/hadoop

nevillelyh:


davorbonaci:
@tomwhite Tom, as an original writer, do you have any comments on this pull request? If not, we'll do a quick review and merge @nevillelyh's changes.

contrib/hadoop/src/main/java/com/google/cloud/dataflow/contrib/hadoop/AvroHadoopFileSource.java
```diff
@@ -0,0 +1,86 @@
+package com.google.cloud.dataflow.contrib.hadoop;
```
davorbonaci: We require a copyright header in every file. Could you add it please?

contrib/hadoop/src/main/java/com/google/cloud/dataflow/contrib/hadoop/AvroWrapperCoder.java
```diff
@@ -0,0 +1,83 @@
...
+
+  @Override
+  public CloudObject asCloudObject() {
+    CloudObject result = super.asCloudObject();
+    result.put("wrapperType", wrapperType.getName());
+    return result;
+  }
+
+  @Override
+  public void verifyDeterministic() throws NonDeterministicException {}
```
davorbonaci: Is this really a deterministic coder?

contrib/hadoop/src/main/java/com/google/cloud/dataflow/contrib/hadoop/HadoopFileSink.java
```diff
@@ -0,0 +1,211 @@
+package com.google.cloud.dataflow.contrib.hadoop;
```
davorbonaci: Copyright header is needed here.

davorbonaci:
I put in a few comments. The main comment would be to add more tests. It is often easy to test coders using `CoderProperties` utility. That said, we probably wouldn't block merging this pull request on it.

nevillelyh:
@davorbonaci I addressed most of your comments. The only thing missing is tests for the new source/sink.

davorbonaci:
@nevillelyh We are about ready to take this pull request, but we'd like to take it directly in our Apache Beam repository. Would you mind resubmitting there?

Note pending apache/incubator-beam#96, which does some renaming. If possible, please apply this fix on top of that change.

And, thanks for this! Much appreciated.

PR #108, opened 2016-02-04
Change BQ mode to append instead of truncate

ravwojdyla:
It appears that truncate mode is not supported for unbounded
PCollection.

Error message:

[ERROR] Failed to execute goal
org.codehaus.mojo:exec-maven-plugin:1.1:java (default-cli) on
project
google-cloud-dataflow-java-examples-all: An exception occured while
executing th
e Java class. null: InvocationTargetException:
WriteDisposition.WRITE_TRUNCATE is not supported for unbounded
PCollections or when using tablespec functions. -> [Help 1]
[ERROR]
[ERROR] To see the full stack trace of the errors, re-run Maven with
the
-e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug
logging.
[ERROR]
[ERROR] For more information about the errors and possible
solutions,
please read the following articles:
[ERROR] [Help 1]
http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException

PR #110, opened 2016-02-10
Disable Maven forking in Travis-CI to fix build issues

dhalperi:
* See travis-ci/travis-ci#3396. Basically, there's an error that does not properly pass -Xmx and similar arguments to the forked JVMs that Javadoc, Surefire, and other plugins invoke. This bug causes the forked JVMs to exceed the container's RAM limit, the JVMs to get killed, and the builds to fail.

    Disabling the forking by setting `-DforkCount=0` in the Maven command seems to have resolved the issue, and it does not slow down build time much.

* However, the underlying bug seems to only affect Linux VMs, and not OS X. Additionally, disabling forking in the OS X tests causes them to fail for another lingering Travis issue (PermSize in OS X). So, I rewrote the test matrix to only disable forking with the above argument for Linux tests.

* In the cleanup, I rewrote the test matrix for Travis to be simpler and cleaner.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,29 +9,18 @@ notifications:
     on_success: change
     on_failure: always
 
-os:
-  - linux
-  - osx
-
-env:
-  matrix:
-    - CUSTOM_JDK="default"
-    - CUSTOM_JDK="oraclejdk8"
-    - CUSTOM_JDK="oraclejdk7"
-    - CUSTOM_JDK="openjdk7"
-
 matrix:
-  exclude:
-     # On OSX, run with default JDK only.
-     - os: osx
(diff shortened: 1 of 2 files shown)


dhalperi:
CLA for Rafal is confirmed via Spotify, as in PR #108.

This PR includes #108 so that I could fix the build issues.

dhalperi:
Ugh, now causing OS X build problems. Unassigning

dhalperi:
Problems fixed; involved a test matrix rewrite. Though the new matrix is much cleaner. See the updated PR description.

davorbonaci:
LGTM. Merging.

PR #111, opened 2016-02-10
Proposing changes to support HBase Sequence Files.

weiminyu:
Proposing following changes to HadoopFileSource:
1. Allow user to override the output coder. 
2. Allow user to pass configuration properties to the file Reader. For example, hbase.import.version, 
   which decides  what version of Result's Deserializer is used. User may also choose to pass 
   gcs-connector configs this way  instead of putting them in core-site.xml
3. Allow user to turn off input file access by the source from the starting host. See HadoopFileSource#setIsRemoteFileFromLaunchSite() for motivation.

A copy of this HadoopFileSource with the proposed changes is used here:
https://github.com/GoogleCloudPlatform/cloud-bigtable-client/tree/master/bigtable-dataflow-import/src/main/java/com/google/cloud/bigtable/dataflowimport

### DIFF
--- a/contrib/hadoop/pom.xml
+++ b/contrib/hadoop/pom.xml
@@ -165,5 +165,11 @@
       <version>4.11</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.mockito</groupId>
+      <artifactId>mockito-core</artifactId>
+      <version>1.10.19</version>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 </project>
--- a/contrib/hadoop/src/main/java/com/google/cloud/dataflow/contrib/hadoop/HadoopFileSource.java
+++ b/contrib/hadoop/src/main/java/com/google/cloud/dataflow/contrib/hadoop/HadoopFileSource.java
@@ -23,9 +23,11 @@ import com.google.cloud.dataflow.sdk.io.BoundedSource;
 import com.google.cloud.dataflow.sdk.io.Read;
 import com.google.cloud.dataflow.sdk.options.PipelineOptions;
 import com.google.cloud.dataflow.sdk.values.KV;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Function;
 import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Lists;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileStatus;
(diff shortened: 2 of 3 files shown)


davorbonaci:
Hi Weimin,
Adding additional configuration properties to the Reader sounds like a nice improvement.

Overriding a coder can often be done differently. For example, could the same thing be achieved via `.apply(Read.from(anySource)).setCoder(anyCoder)`? If not, can you please elaborate more.

Regarding suppression of warnings during job submission, perhaps this could be done differently. Sounds problematic that the user needs to set this via a method (which is also static). Could the logging be disabled elsewhere?

Finally, the continuous integration failed due to an issue in our scripts. If you rebase, that issue should go away.

Thanks!

weiminyu:
Hi Davor, I responded in line:


On Thu, Feb 11, 2016 at 9:45 PM, Davor Bonaci <notifications@github.com>
wrote:

> Hi Weimin,
> Adding additional configuration properties to the Reader sounds like a
> nice improvement.
>
> Overriding a coder can often be done differently. For example, could the
> same thing be achieved via .apply(Read.from(anySource)).setCoder(anyCoder)?
> If not, can you please elaborate more.
>
Using setCoder() does not work because the source's default coder is always
requested during initialization in Read.Bounded. The relevant part of stack
trace is as follows:
at
com.google.cloud.bigtable.dataflowimport.HadoopFileSource.getDefaultOutputCoder(HadoopFileSource.java:288)
at
com.google.cloud.dataflow.sdk.io.Read$Bounded.getDefaultOutputCoder(Read.java:119)
at com.google.cloud.dataflow.sdk.io.Read$Bounded.apply(Read.java:128)
at com.google.cloud.dataflow.sdk.io.Read$Bounded.apply(Read.java:99)
at
com.google.cloud.dataflow.sdk.runners.PipelineRunner.apply(PipelineRunner.java:74)
at
com.google.cloud.dataflow.sdk.runners.DirectPipelineRunner.apply(DirectPipelineRunner.java:236)
at com.google.cloud.dataflow.sdk.Pipeline.applyInternal(Pipeline.java:367)
at com.google.cloud.dataflow.sdk.Pipeline.applyTransform(Pipeline.java:290)
at com.google.cloud.dataflow.sdk.values.PBegin.apply(PBegin.java:58)
at com.google.cloud.dataflow.sdk.Pipeline.apply(Pipeline.java:171)

Are there other ways?


> Regarding suppression of warnings during job submission, perhaps this
> could be done differently. Sounds problematic that the user needs to set
> this via a method (which is also static). Could the logging be disabled
> elsewhere?
>
Disabling logging through logger configuration has the same problem: we
want to disable logging on my desktop but not on the cloud where an access
failure is a real problem.  We also don't want to disable logging if the
job is run locally using DirectPipelineRunner. These are conditions
that the source cannot figure out, and seems only a problem with the
hadoop-gcs combination.

The reason why the log-suppressing flag is static is that the pipeline
serializes-then-deserializes the source before analyzing it. So a transient
field is not enough.


> Finally, the continuous integration failed due to an issue in our scripts.
> If you rebase, that issue should go away.
>
Thanks, and done.

dhalperi:
Closing as obsolete

PR #115, opened 2016-02-17
BEAM-13 - Create JMS IO

jbonofre:
Work in progress to provide a JMS IO to Dataflow, allowing to consume and produce messages from a JMS queue or topic.

### DIFF
--- a/contrib/jms/README.md
+++ b/contrib/jms/README.md
@@ -1,0 +1,45 @@
+# JMS module
+
+This library provides Dataflow sources and sinkgs to make it possible to read
+and write on JMS brokers from Dataflow pipelines.
+
+It supports both JMS queues and topics, with unbounded or bounded `PCollections`.
+
+To use JmsIO, you have to:
+
+1. Create a JMS `ConnectionFactory` specific to the JMS broker you want to use (for instance, ActiveMQ, IBM MQ, ...)
+2. Specify the JMS destination (queue or topic) you want to use
+
+## Reading (consuming messages) with JmsIO
+
+The `JmsIO.Read` transform continuously consumes from the JMS broker and returns an unbounded `PCollection` of `Strings` that
+represent the messages. By default, each element in the resulting `PCollection` is encoded as a UTF-8 string.
+You can override the default encoding by using `withCoder` when you call JmsIO.Read.
+
+----
+PipelineOptions options = PipelineOptionsFactory.create();
+Pipeline p = Pipeline.create(options);
+
+// create JMS connection factory, for instance, with ActiveMQ
+javax.jms.ConnectionFactory connectionFactory = new ActiveMQConnectionFactory("tcp://localhost:61616");
+
+PCollection<String> data = p.apply(JmsIO.Read.named("ConsumeFromJMS").connectionFactory(connectionFactory).queue("my-queue"));
+----
+
+### Reading a bounded set of messages from JMS
+
+The `DirectPipelineRunner` and the batch mode of the Dataflow service do not support unbounded `PCollections`.
+To use JmsIO as source in these contexts, you need to supply a bound on the amount of messages to consume.
+
+You can specify the `.maxNumMessages` option to read a fixed maximum number of messages.
+
+## Writing (producing messages) with JmsIO
+
+The JmsIO.Write transform continuously writes an unbounded `PCollection` of `String` objects, produced to a
+JMS broker. By default, the input `PCollection` to `JmsIO.Write` must contain strings encoded in UTF-8.
+You can change the expected input type and encoding by using `withCoder`.
+
+----
+PCollection<String> data = ...;
+data.apply(JmsIO.Write.named("ProduceToJMS").connectionFactory(connectionFactory).queue("my-queue"));
+----
--- a/contrib/jms/src/main/java/com/google/cloud/dataflow/contrib/jms/JmsIO.java
+++ b/contrib/jms/src/main/java/com/google/cloud/dataflow/contrib/jms/JmsIO.java
@@ -1,0 +1,532 @@
+/*
+ * Copyright (C) 2015 Google Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+package com.google.cloud.dataflow.contrib.jms;
+
+import com.google.cloud.dataflow.sdk.coders.Coder;
+import com.google.cloud.dataflow.sdk.coders.StringUtf8Coder;
+import com.google.cloud.dataflow.sdk.coders.VoidCoder;
+import com.google.cloud.dataflow.sdk.transforms.Create;
+import com.google.cloud.dataflow.sdk.transforms.DoFn;
+import com.google.cloud.dataflow.sdk.transforms.PTransform;
+import com.google.cloud.dataflow.sdk.transforms.ParDo;
+import com.google.cloud.dataflow.sdk.util.CoderUtils;
+import com.google.cloud.dataflow.sdk.util.WindowingStrategy;
+import com.google.cloud.dataflow.sdk.values.PCollection;
+import com.google.cloud.dataflow.sdk.values.PDone;
+import com.google.cloud.dataflow.sdk.values.PInput;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import javax.annotation.Nullable;
+
+import javax.jms.Connection;
+import javax.jms.ConnectionFactory;
+import javax.jms.MessageConsumer;
+import javax.jms.MessageProducer;
+import javax.jms.Session;
+import javax.jms.TextMessage;
+
+/**
+ * Read and Write {@link PTransform}s for JMS broker. These transforms create and consume
+ * unbounded {@link PCollection PCollections}.
+ */
+public class JmsIO {
+
+    private static final Logger LOG = LoggerFactory.getLogger(JmsIO.class);
+
+    /** The default {@link Coder} used to translate to/from JMS messages. */
+    public static final Coder<String> DEFAULT_JMS_CODER = StringUtf8Coder.of();
+
+    /**
+     * A {@link PTransform} that continuously reads from a JMS broker and
+     * return a {@link PCollection} of {@link String Strings} containing the items.
+     */
+    public static class Read {
+
+        /**
+         * Creates and returns a transform for reading from the JMS broker with the specified
+         * transform name.
+         *
+         * @param name the step name
+         * @return the Bound
+         */
+        public static Bound<String> named(String name) {
+            return new Bound<>(DEFAULT_JMS_CODER).named(name);
+        }
+
+        /**
+         * Creates and returns a transform for reading from a JMS broker with the provided
+         * JMS connection factory.
+         *
+         * @param connectionFactory the JMS connection factory to use to connect to the broker
+         * @return the Bound
+         */
+        public static Bound<String> connectionFactory(ConnectionFactory connectionFactory) {
+            return new Bound<>(DEFAULT_JMS_CODER).connectionFactory(connectionFactory);
+        }
+
+        /**
+         * Creates and returns a transform for reading from a JMS broker with the provided
+         * JMS queue.
+         *
+         * @param name the JMS queue name
+         * @return the Bound
+         */
+        public static Bound<String> queue(String name) {
+            return new Bound<>(DEFAULT_JMS_CODER).queue(name);
+        }
+
+        /**
+         * Creates and returns a transform for reading from a JMS broker with the provided
+         * JMS topic.
+         *
+         * @param name the JMS topic name.
+         * @return the Bound
+         */
+        public static Bound<String> topic(String name) {
+            return new Bound<>(DEFAULT_JMS_CODER).topic(name);
+        }
+
+        /**
+         * Creates and returns a transform for reading from JMS broker that uses the given
+         * {@link Coder} to decode JMS messages into a value of type {@code T}.
+         *
+         * @param coder the coder to use
+         * @return the Bound
+         */
+        public static <T> Bound<T> withCoder(Coder<T> coder) {
+            return new Bound<>(coder);
+        }
+
+        /**
+         * Creates and returns a tranform for reading from JMS broker that limit the
+         * number of messages consumed.
+         *
+         * @param maxNumMessages the max number of messages consumed
+         * @return the Bound
+         */
+        public static Bound<String> maxNumMessages(int maxNumMessages) {
+            return new Bound<>(DEFAULT_JMS_CODER).maxNumMessages(maxNumMessages);
+        }
+
+        /**
+         * A {@link PTransform} that reads from a JMS broker and returns
+         * a unbounded {@link PCollection} containing the items from the stream.
+         */
+        public static class Bound<T> extends PTransform<PInput, PCollection<T>> {
+
+            /** The JMS connection factory. */
+            private final ConnectionFactory connectionFactory;
+
+            /** The JMS queue to read from. */
+            private final String queue;
+
+            /** The JMS topic to read from. */
+            private final String topic;
+
+            /** The coder used to decode each message. */
+            @Nullable
+            private final Coder<T> coder;
+
+            /** Stop after reading this many messages. */
+            private final int maxNumMessages;
+
+            private Bound(Coder<T> coder) {
+                this(null, null, null, null, coder, 0);
+            }
+
+            private Bound(String name, ConnectionFactory connectionFactory, String queue,
+                          String topic, Coder<T> coder, int maxNumMessages) {
+                super(name);
+                this.connectionFactory = connectionFactory;
+                this.queue = queue;
+                this.topic = topic;
+                this.coder = coder;
+                this.maxNumMessages = maxNumMessages;
+            }
+
+            /**
+             * Returns a transform that's like this one but with the given step name.
+             *
+             * @param name the step name
+             * @return the Bound
+             */
+            public Bound<T> named(String name) {
+                return new Bound<T>(name, connectionFactory, queue, topic, coder, maxNumMessages);
+            }
+
+            /**
+             * Returns a transform that's like this one but using the JMS connection factory.
+             *
+             * @param connectionFactory the JMS connection factory to use
+             * @return the Bound
+             */
+            public Bound<T> connectionFactory(ConnectionFactory connectionFactory) {
+                return new Bound<T>(name, connectionFactory, queue, topic, coder, maxNumMessages);
+            }
+
+            /**
+             * Returns a transform that's like this one but that reads
+             * from the specified JMS queue.
+             *
+             * @param queue the JMS queue
+             * @return the Bound
+             */
+            public Bound<T> queue(String queue) {
+                return new Bound<T>(name, connectionFactory, queue, topic, coder, maxNumMessages);
+            }
+
+            /**
+             * Returns a transform that's like this one but that reads
+             * from the specified JMS topic.
+             *
+             * @param topic the JMQ topic
+             * @return the Bound
+             */
+            public Bound<T> topic(String topic) {
+                return new Bound<T>(name, connectionFactory, queue, topic, coder, maxNumMessages);
+            }
+
+            /**
+             * Returns a transform that's like this one but that uses the
+             * given {@link Coder} to decode each record into a value of
+             * type {@code X}.
+             *
+             * @param coder the coder to use
+             * @return the Bound
+             */
+            public <X> Bound<X> withCoder(Coder<X> coder) {
+                return new Bound<>(name, connectionFactory, queue, topic, coder, maxNumMessages);
+            }
+
+            /**
+             * Return a transform that's like this one but will only read up
+             * to the specified maximum number of messages from the JMS broker.
+             * The transform produces a <i>bounded</i> {@link PCollection}.
+             *
+             * @param maxNumMessages the max number messages limit to consume
+             * @return the Bound
+             */
+            public Bound<T> maxNumMessages(int maxNumMessages) {
+                return new Bound<T>(name, connectionFactory, queue, topic, coder, maxNumMessages);
+            }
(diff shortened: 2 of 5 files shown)


jbonofre:
I signed it !

On 02/17/2016 02:47 PM, googlebot wrote:
> Thanks for your pull request. It looks like this may be your first
> contribution to a Google open source project. Before we can look at your
> pull request, you'll need to sign a Contributor License Agreement (CLA).
>
> :memo: *Please visit https://cla.developers.google.com/ to sign.*
>
> Once you've signed, please reply here (e.g. |I signed it!|) and we'll
> verify. Thanks.
>
> ------------------------------------------------------------------------
>
>   * If you've already signed a CLA, it's possible we don't have your
>     GitHub username or you're using a different email address. Check
>     your existing CLA data <https://cla.developers.google.com/clas> and
>     verify that your email is set on your git commits
>     <https://help.github.com/articles/setting-your-email-in-git/>.
>   * If you signed the CLA as a corporation, please let us know the
>     company's name.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/GoogleCloudPlatform/DataflowJavaSDK/pull/115#issuecomment-185210226>.
>

-- 
Jean-Baptiste Onofré
jbonofre@apache.org
http://blog.nanthrax.net
Talend - http://www.talend.com

contrib/jms/README.md
```diff
@@ -0,0 +1,45 @@
+# JMS module
+
+This library provides Dataflow sources and sinkgs to make it possible to read
```
dhalperi: sinks

contrib/jms/README.md
```diff
@@ -0,0 +1,45 @@
+# JMS module
```
dhalperi: probably you want to say what JMS is and provide a link somewhere in the first line or two of the README

contrib/jms/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+This library provides Dataflow sources and sinkgs to make it possible to read
+and write on JMS brokers from Dataflow pipelines.
+
+It supports both JMS queues and topics, with unbounded or bounded `PCollections`.
```
dhalperi: merge into previous paragraph?

dhalperi:
Hi JB,

Under the assumption that as a core committer you want a "lot" of review, I'll try to overdo giving you comments. Feel free to ignore (with a comment) ones that you feel aren't appropriate.

Thanks!
Dan

jbonofre:
Awesome ! Thanks a lot, much appreciate. I take all comments !!

contrib/jms/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+PCollection<String> data = p.apply(JmsIO.Read.named("ConsumeFromJMS").connectionFactory(connectionFactory).queue("my-queue"));
+----
+
+### Reading a bounded set of messages from JMS
+
+The `DirectPipelineRunner` and the batch mode of the Dataflow service do not support unbounded `PCollections`.
+To use JmsIO as source in these contexts, you need to supply a bound on the amount of messages to consume.
+
+You can specify the `.maxNumMessages` option to read a fixed maximum number of messages.
```
dhalperi: I find it interesting that you separated out the above two sections. Is there a good reason for this?

I would have kept it simpler, and more unified, something like this:

## Reading from JMS
The `JmsIO.Read` transform produces an unbounded `PCollection` containing messages received from a JMS queue. By default, each element in the output collection is a `UTF-8`-encoded string. You can override the default encoding using `JmsIO.Read.withCoder`.

```java
Pipeline p = ...;

ConnectionFactory connectionFactory = new ActiveMQConnectionFactory("tcp://localhost:61616");

// Producing Strings by default
JmsIO.Read read = JmsIO.Read.connectionFactory(connectionFactory).queue("my-queue");
PCollection<String> data = p.apply("Read from ActiveMQ", read);

// Producing a byte[] using a different Coder
JmsIO.Read read = JmsIO.Read.connectionFactory(connectionFactory).queue("my-queue").withCoder(ByteArrayCoder.class);
```

`JmsIO` can also produce a *bounded* `PCollection` by limiting either the number of records returned (`Read.withMaxNumRecords`) or the time for which records are consumed (`Read.withMaxReadTime`). Note that if only a record limit is configured, your pipeline will not finish executing until the desired number of records is received.

```java
JmsIO.Read read = JmsIO.Read.connectionFactory(connectionFactory).queue("my-queue").withMaxNumRecords(5);
```

====

There should probably be a type variable on `JmsIO.Read`, like `JmsIO.Read<String>`, `JmsIO.Read<byte[]>`, but I haven't gotten to that code yet.

contrib/jms/README.md
```diff
@@ -0,0 +1,45 @@
...
+// create JMS connection factory, for instance, with ActiveMQ
+javax.jms.ConnectionFactory connectionFactory = new ActiveMQConnectionFactory("tcp://localhost:61616");
+
+PCollection<String> data = p.apply(JmsIO.Read.named("ConsumeFromJMS").connectionFactory(connectionFactory).queue("my-queue"));
+----
+
+### Reading a bounded set of messages from JMS
+
+The `DirectPipelineRunner` and the batch mode of the Dataflow service do not support unbounded `PCollections`.
+To use JmsIO as source in these contexts, you need to supply a bound on the amount of messages to consume.
```
dhalperi: This text makes me think that bounded use of JmsIO is a weird corner case. I think there's plenty of reasons a user may want to operate in bounded mode, e.g., for testing or during early pipeline development, or because they're loading historical data. So the text below de-marginalizes it.

contrib/jms/README.md
dhalperi: I would keep the unbounded -> bounded method names unchanged from [`BoundedReadFromUnboundedSource`](https://github.com/GoogleCloudPlatform/DataflowJavaSDK/blob/master/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedReadFromUnboundedSource.java#L61): `withMaxNumRecords` and `withMaxReadTime`.

More unified API.

I'd also suggest copying both methods, not just limiting records, because the blocking semantics or limiting records only has tripped users up in the past.

contrib/jms/README.md
```diff
@@ -0,0 +1,45 @@
...
+You can override the default encoding by using `withCoder` when you call JmsIO.Read.
+
+----
+PipelineOptions options = PipelineOptionsFactory.create();
+Pipeline p = Pipeline.create(options);
+
+// create JMS connection factory, for instance, with ActiveMQ
+javax.jms.ConnectionFactory connectionFactory = new ActiveMQConnectionFactory("tcp://localhost:61616");
+
+PCollection<String> data = p.apply(JmsIO.Read.named("ConsumeFromJMS").connectionFactory(connectionFactory).queue("my-queue"));
```
dhalperi: drop the `.named` code, in favor of just applying the name during `p.apply`. I think its just unnecessary code in this case.

contrib/jms/README.md
```diff
@@ -0,0 +1,45 @@
...
+----
+
+### Reading a bounded set of messages from JMS
+
+The `DirectPipelineRunner` and the batch mode of the Dataflow service do not support unbounded `PCollections`.
+To use JmsIO as source in these contexts, you need to supply a bound on the amount of messages to consume.
+
+You can specify the `.maxNumMessages` option to read a fixed maximum number of messages.
+
+## Writing (producing messages) with JmsIO
```
dhalperi: Similar changes here.

* `.connectionFactory` -> `.withConnectionFactory`.
* The section could be more concisely "Writing to JMS"
* drop "continuously".
* hopefully, the writer takes a bounded or unbounded PCollection, not just an unbounded PCollection.

contrib/jms/src/main/java/com/google/cloud/dataflow/contrib/jms/JmsIO.java
```diff
@@ -0,0 +1,532 @@
...
+import javax.jms.ConnectionFactory;
+import javax.jms.MessageConsumer;
+import javax.jms.MessageProducer;
+import javax.jms.Session;
+import javax.jms.TextMessage;
+
+/**
+ * Read and Write {@link PTransform}s for JMS broker. These transforms create and consume
+ * unbounded {@link PCollection PCollections}.
+ */
```
dhalperi: We tend to put the same type of code you wrote in the README here as well.

For example, see,  [`CountingSource`](https://github.com/GoogleCloudPlatform/DataflowJavaSDK/blob/master/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/CountingSource.java#L37)

contrib/jms/src/main/java/com/google/cloud/dataflow/contrib/jms/JmsIO.java
```diff
@@ -0,0 +1,532 @@
...
+public class JmsIO {
+
+    private static final Logger LOG = LoggerFactory.getLogger(JmsIO.class);
```
dhalperi: Have to stop reviewing here for now. More comments later!

contrib/jms/src/main/java/com/google/cloud/dataflow/contrib/jms/JmsIO.java
```diff
@@ -0,0 +1,532 @@
...
+    public static class Read {
+
+        /**
+         * Creates and returns a transform for reading from the JMS broker with the specified
+         * transform name.
+         *
+         * @param name the step name
+         * @return the Bound
+         */
+        public static Bound<String> named(String name) {
```
dhalperi: per our discussion on the mailing list -- I would drop the `Bounded` class and instead just make `Read` itself be a `PTransform<PBegin, PCollection<T>>`. The name here is unnecessary -- it can be more easily set directly via `p.apply("name", JmsIO.Read...);`

contrib/jms/src/main/java/com/google/cloud/dataflow/contrib/jms/JmsIO.java
```diff
@@ -0,0 +1,532 @@
...
+                return new Bound<T>(name, connectionFactory, queue, topic, coder, maxNumMessages);
+            }
+
+            /**
+             * Returns a transform that's like this one but using the JMS connection factory.
+             *
+             * @param connectionFactory the JMS connection factory to use
+             * @return the Bound
+             */
+            public Bound<T> connectionFactory(ConnectionFactory connectionFactory) {
```
dhalperi: we tend to call these `.withConnectionFactory` instead of `.connectionFactory`.

dhalperi: and similar below.

contrib/jms/src/main/java/com/google/cloud/dataflow/contrib/jms/JmsIO.java
```diff
@@ -0,0 +1,532 @@
...
+                return new Bound<T>(name, connectionFactory, queue, topic, coder, maxNumMessages);
+            }
+
+            /**
+             * Returns a transform that's like this one but that reads
+             * from the specified JMS queue.
```
dhalperi: wrapping seems a little short on many of these javadoc comments.

dhalperi: (reminder that I promised to overdo the comments in this review ;)

contrib/jms/src/main/java/com/google/cloud/dataflow/contrib/jms/JmsIO.java
```diff
@@ -0,0 +1,532 @@
...
+             * @param maxNumMessages the max number messages limit to consume
+             * @return the Bound
+             */
+            public Bound<T> maxNumMessages(int maxNumMessages) {
+                return new Bound<T>(name, connectionFactory, queue, topic, coder, maxNumMessages);
+            }
+
+            @Override
+            public PCollection<T> apply(PInput input) {
+                if (connectionFactory == null) {
```
dhalperi: you can put this code in [`PTransform#validate`](https://github.com/GoogleCloudPlatform/DataflowJavaSDK/blob/master/sdk/src/main/java/com/google/cloud/dataflow/sdk/transforms/PTransform.java#L195) instead.

And since you link in guava, you can feel free to do

```java
import static ...Preconditions.checkNotNull;
...
checkNotNull(connectionFactory, "message here");
```

to simplify.

contrib/jms/src/main/java/com/google/cloud/dataflow/contrib/jms/JmsIO.java
```diff
@@ -0,0 +1,532 @@
...
+                    throw new IllegalArgumentException("need to set connection factory " +
+                            "for a JmsIO.Read transform");
+                }
+                if (queue == null && topic == null) {
+                    throw new java.lang.IllegalStateException("need to set destination " +
```
dhalperi: checkArgument/checkState

contrib/jms/src/main/java/com/google/cloud/dataflow/contrib/jms/JmsIO.java
```diff
@@ -0,0 +1,532 @@
...
+                            "(queue or topic) for a JmsIO.Read transform");
+                }
+
+                boolean boundedOutput = getMaxNumMessages() > 0;
+
+                if (boundedOutput) {
+                    return input.getPipeline().begin()
+                            .apply(Create.of((Void) null)).setCoder(VoidCoder.of())
```
dhalperi: We recommend using [`Create#withCoder`](https://github.com/GoogleCloudPlatform/DataflowJavaSDK/blob/master/sdk/src/main/java/com/google/cloud/dataflow/sdk/transforms/Create.java#L70) instead of setting the coder on the output PCollection.

The reasoning is a little subtle -- any PTransform that changes the type of its input at runtime needs a strong handle on the TypeDescriptor (or at least the Coder) that a user may want to use. This is because the runner-specific implementation of `Create` may be a composite PTransform that needs the coder in `apply`. Note that the output PCollection is created during apply time, so naturally the output coder could not be set yet.

contrib/jms/src/main/java/com/google/cloud/dataflow/contrib/jms/JmsIO.java
```diff
@@ -0,0 +1,532 @@
...
+                            .apply(ParDo.of(new JmsReader())).setCoder(coder);
```
dhalperi: Hmm -- I expected to see you using both the `BoundedSource` and `UnboundedSource` APIs, rather than this approach. Specifically, this approach looks like it won't scale to more than 1 worker.

It looks like you based this implementation on `PubSubIO`? If so, that is very unfortunate -- the version of pub/sub that is in the SDK is not meant for production use, but rather only for testing with the DirectPipelineRunner.  https://issues.apache.org/jira/browse/BEAM-53

contrib/jms/src/main/java/com/google/cloud/dataflow/contrib/jms/JmsIO.java
dhalperi: (I now assume this is because of using PubSUbIO, which does not follow our most modern conventions.)

contrib/jms/src/main/java/com/google/cloud/dataflow/contrib/jms/JmsIO.java
```diff
@@ -0,0 +1,532 @@
...
+     */
+    public static class Write {
+
+        /**
+         * Creates a transform that produces to a JMS broker with the given step name.
+         *
+         * @param name the step name
+         * @return the Bound
+         */
+        public static Bound<String> named(String name) {
```
dhalperi: similar recommendations w.r.t dropping `Write.Bound` and just using `Write` instead, dropping `named`, adding `withX` as appropriate.

contrib/jms/src/main/java/com/google/cloud/dataflow/contrib/jms/JmsIO.java
```diff
@@ -0,0 +1,532 @@
...
+                return new Bound<T>(name, connectionFactory, queue, topic, coder);
+            }
+
+            public <X> Bound<X> withCoder(Coder<X> coder) {
+                return new Bound<>(name, connectionFactory, queue, topic, coder);
+            }
+
+            @Override
+            public PDone apply(PCollection<T> input) {
+                if (connectionFactory == null) {
```
dhalperi: in `#validate`?

contrib/jms/src/main/java/com/google/cloud/dataflow/contrib/jms/JmsIO.java
```diff
@@ -0,0 +1,532 @@
...
+                return coder;
+            }
+
+            private class JmsWriter extends DoFn<T, Void> {
+                private transient Connection connection;
+                private transient Session session;
+
+                @Override
+                public void startBundle(Context c) throws Exception {
+                    connection = connectionFactory.createConnection();
```
dhalperi: A few motivating questions:

* do you expect to use this Write for batch jobs or streaming jobs or both?
* how expensive is it to create a connection?
* what resources does each connection use, and are they likely to run into machine-local resource limits?

In batch pipelines, there may be few, very large bundles. So it makes sense to open a connection at the beginning and close it at the end.

However, in streaming pipelines, there are many small, short-lived bundles (very few elements, even 1, per bundle!). If opening a connection takes a long time, this can add a lot of overhead to a pipeline. Additionally, there can be several hundred threads each with an active bundle (so, several hundred connections open) so it's important to think about resource limits.

contrib/jms/src/main/java/com/google/cloud/dataflow/contrib/jms/JmsIO.java
```diff
@@ -0,0 +1,532 @@
...
+                        producer = session.createProducer(session.createTopic(topic));
+                    }
+                    producer.send(message);
+                    producer.close();
+                }
+
+                @Override
+                public void finishBundle(Context c) throws Exception {
+                    if (session != null) {
+                        session.close();
```
dhalperi: Is it important to make sure that `connection` is also closed even if `session.close()` throws an exception?

contrib/jms/src/test/java/com/google/cloud/dataflow/contrib/jms/JmsIOTest.java
```diff
@@ -0,0 +1,140 @@
...
+
+import javax.jms.Connection;
+import javax.jms.ConnectionFactory;
+import javax.jms.MessageConsumer;
+import javax.jms.MessageProducer;
+import javax.jms.Session;
+import javax.jms.TextMessage;
+
+/**
+ * Test of the JmsIO.
```
dhalperi: `{@link JmsIO}`? Survives across refactorings, etc.

contrib/jms/src/test/java/com/google/cloud/dataflow/contrib/jms/JmsIOTest.java
```diff
@@ -0,0 +1,140 @@
...
+public class JmsIOTest {
+
+    private static final String BROKER_URL = "vm://localhost";
+
+    private BrokerService broker;
+    private ConnectionFactory connectionFactory;
+
+    @Before
+    public void startBroker() throws Exception {
+        System.out.println("Starting ActiveMQ broker on " + BROKER_URL);
```
dhalperi: use `java.slf4j.Logger` rather than `stderr`/`stdout`?

contrib/jms/src/test/java/com/google/cloud/dataflow/contrib/jms/JmsIOTest.java
```diff
@@ -0,0 +1,140 @@
...
+                JmsIO.Read
+                        .connectionFactory(connectionFactory)
+                        .queue("test")
+                        .maxNumMessages(5));
+
+        System.out.println("Starting the pipeline");
+        pipeline.run();
+
+        System.out.println("Check assertion");
+        DataflowAssert.that(output).containsInAnyOrder(new String[]{ "test "});
```
dhalperi: Hmm. I would have expected a test more like

```java
List<String> testMessages = ImmutableList.of("msg1", "msg2", "msg3");
...
for (String m : testMessages) {
    producer.send(session.createTextMessage(m));
}
...

...
Pipeline pipeline = ...;
PCollection<String> output = pipeline.apply(...);
DataflowAssert.that(output).containsInAnyOrder(testMessages);
pipeline.run();
```

Note that the `DataflowAssert.that()...` line must go **before** the pipeline is run. Otherwise it's not actually in the pipelien that is executed!

Note that this test should fail -- `test` is not the test message that you sent! -- but the assert is not being run.

dhalperi: filed https://issues.apache.org/jira/browse/BEAM-85

contrib/jms/src/test/java/com/google/cloud/dataflow/contrib/jms/JmsIOTest.java
```diff
@@ -0,0 +1,140 @@
...
+        MessageConsumer messageConsumer = session.createConsumer(session.createQueue("test"));
+        TextMessage message = (TextMessage) messageConsumer.receive(10000);
+        Assert.assertEquals(message.getText(), "Test");
+    }
+
+    @After
+    public void stopBroker() throws Exception {
+        broker.stop();
+    }
+
```
dhalperi: I would like to see a few more unit tests in this file. Things like building the Read and Write transform works, and that validation catches invalid Read and Write transforms. For examples, maybe see https://github.com/GoogleCloudPlatform/DataflowJavaSDK/blob/master/sdk/src/test/java/com/google/cloud/dataflow/sdk/io/bigtable/BigtableIOTest.java#L159 and related tests.

Basically, improving test coverage of the relevant lines in the file. In general, the more coverage the better tho obviously we should be aware of diminishing returns.

contrib/jms/src/test/java/com/google/cloud/dataflow/contrib/jms/JmsIOTest.java
```diff
@@ -0,0 +1,140 @@
...
+        pipeline.run();
+
+        System.out.println("Check assertion");
+        DataflowAssert.that(output).containsInAnyOrder(new String[]{ "test "});
+
+        System.out.println(output);
+    }
+
+    @Test
+    public void testWriteMessage() throws Exception {
```
dhalperi: I like that you test read and write separately. Specifically, it's nice to use JMS write & JmsIO.Read, plus standard JMS read with JmsIO.Write.

dhalperi: (no action needed, just commenting)

dhalperi:
Hi JB,

I know you are still making changes here based on our last discussions. When you're ready, would you prefer to continue here or to make a PR against the main incubator-beam repo? If the latter, let's close this and reopen once you're ready.

Thanks.
Dan

jbonofre:
Hi Dan,

I will create a new PR at Beam. Let me close this PR then.

Thanks !
Regards
JB

PR #118, opened 2016-02-17
Make sure every hold is accompanied by a proximate timer

mshields822:
We now maintain the invariant that every hold
inside(outside) a window is accompanied by a
timer inside(outside) the window.


mshields822:
I signed it!

mshields822:
Try again.

PR #121, opened 2016-02-19
[BEAM-52] Kafka custom source 

rangadi:
Current status (3/18/16) : well tested and feature complete. going through reviews.

--
This is a preliminary PR. It is not tested yet. I am working on example application and unit tests. Some TODOs : 

- [x] unit tests
- [x] testing in the cluster with larger volumes of data and larger number of partitions
- [x] add more stats (I need to look into stats support in sdk)


rangadi:
I signed the CLA.

rangadi:
The functionality is well tested. will update javadoc today.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaCheckpointMark.java
```diff
@@ -0,0 +1,64 @@
...
+
+import com.google.cloud.dataflow.sdk.coders.DefaultCoder;
+import com.google.cloud.dataflow.sdk.coders.SerializableCoder;
+import com.google.cloud.dataflow.sdk.io.UnboundedSource;
+
+/**
+ * Checkpoint for an unbounded KafkaSource reader. Consists of Kafka topic name, partition id,
+ * and the latest offset consumed so far.
+ *
+ * @author rangadi
```
dhalperi: We tend not to use these.

rangadi: I like the policy. removed @author tags.

rangadi:
updates:

- move TopHashtagsExample to separate directory hierarchy
- many checkstyle and javadoc fixes 
- invoke watermarkFn with null record if the reader hasn't read any records.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaSource.java
```diff
@@ -0,0 +1,806 @@
+/*
```
dhalperi: Could you rename this file `KafkaIO` and have the `PTransform<PBegin, [output type]>` be `KafkaIO.Read` or something similar? This follows examples like in `BigQueryIO`, `PubSubIO`, `BigtableIO`.

Hiding the source from the caller and exposing PTransforms generally makes the sources more usable and also allows for evolving the implementation over time.

(As an example, I'd like to rewrite `DatastoreIO.Read` based on `Create` instead of `BoundedSource`, but since we exposed the `BoundedSource` it's hard to do in a backwards-compatible way.)

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaSource.java
```diff
@@ -0,0 +1,806 @@
...
+import java.util.NoSuchElementException;
+
+import javax.annotation.Nullable;
+
+/**
+ * Dataflow Source for consuming Kafka sources.
+ *
+ * <pre>
+ * Usage:
+ *        UnboundedSource&lt;String, ?&gt; kafkaSource = KafkaSource
```
dhalperi: I had thought that in our initial discussion the plan was to implement both a Bounded and Unbounded source version.

For now, using `UnboundedSource` in conjunction with `BoundedReadFromUnbounded` source makes sense, but -- if it makes sense -- it will eventually be nicer to do both so that the BoundedSource can also use dynamic work rebalancing.

dhalperi: (and again, exposing a PTransform instead of an UnboundedSource would seem to make this easier -- the code can switch between the two implentations in `.apply`)

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaSource.java
```diff
@@ -0,0 +1,806 @@
...
+ *        pipeline
+ *          .apply(Read.from(kafkaSource).named("read_topic_a_and_b"))
+ *          ....
+ * </pre>
+ */
+public class KafkaSource {
+
+  private static final Logger LOG = LoggerFactory.getLogger(KafkaSource.class);
+
+  private static class NowTimestampFn<T> implements SerializableFunction<T, Instant> {
```
dhalperi: This really ought to be in a TimestampFunctions library,

dhalperi: (that note is more for me than for you.. sorry)

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaSource.java
```diff
@@ -0,0 +1,806 @@
...
+
+  public static Builder<byte[], byte[]> unboundedByteSourceBuilder() {
+    return new Builder<byte[], byte[]>()
+      .withKeyCoder(ByteArrayCoder.of())
+      .withValueCoder(ByteArrayCoder.of());
+  }
+
+  /**
+   * Similar to {@link #unboundedSourceBuilder()}, where user in only interested in value, and
+   * wants to discard Kafak record key and metadata.
```
dhalperi: Kafka

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaSource.java
```diff
@@ -0,0 +1,806 @@
...
+ * <pre>
+ * Usage:
+ *        UnboundedSource&lt;String, ?&gt; kafkaSource = KafkaSource
+ *            .&lt;String&gt;unboundedValueSourceBuilder()
+ *            .withBootstrapServers("broker_1:9092,broker_2:9092)
+ *            .withTopics(ImmutableList.of("topic_a", "topic_b")
+ *            .withValueCoder(StringUtf8Coder.of())
+ *            .withTimestampFn(timestampFn)
+ *            .withWatermarkFn(watermarkFn)
+ *            .build();
```
dhalperi: It seems like there are three different levels of abstraction the user may want from their Kafka reader:

1. `KV<K, V>`  (keys and values)
2. `V`  (values)
3. `KafkaRecord<K, V>`  (keys and values, plus metadata about the record).

based on the Kafka docs I read [here](http://kafka.apache.org/documentation.html#connect_streamsandrecords) I would probably suggest that 1 is the default, and there should be a `Read#withMetadata` builder that turns it into 3.

I'm not sure that 2 is necessary at all -- it's easy to turn a `PCollection<KV<K,V>>` into a `PCollection<V>` using the [`Values` transform](https://github.com/GoogleCloudPlatform/DataflowJavaSDK/blob/master/sdk/src/main/java/com/google/cloud/dataflow/sdk/transforms/Values.java#L58)

Thoughts?

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaSource.java
```diff
@@ -0,0 +1,806 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(KafkaSource.class);
+
+  private static class NowTimestampFn<T> implements SerializableFunction<T, Instant> {
+    @Override
+    public Instant apply(T input) {
+      return Instant.now();
+    }
+  }
+
+  public static <K, V> Builder<K, V> unboundedSourceBuilder() {
```
dhalperi: would think this is just `KafkaIO.read()` static factory method for `KafkaIO.Read` class.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaSource.java
```diff
@@ -0,0 +1,806 @@
...
+    return new Builder<K, V>();
+  }
+
+  public static Builder<byte[], byte[]> unboundedByteSourceBuilder() {
```
dhalperi: Maybe ought to be the default values for `K`, `V` in `KafkaIO.read()`, right?

dhalperi: (or string, and then this can still be constructed using `.withKeyCoder`/`.withValueCoder`

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaSource.java
```diff
@@ -0,0 +1,806 @@
...
+       .withKeyCoder(ByteArrayCoder.of()));
+  }
+
+  /**
+   * Builds Unbounded Kafka Source.
+   *
+   * @param K key type
+   * @param V value type
+   */
+  public static class Builder<K, V> {
```
dhalperi: This is totally ad-libbed, but what about something like this:

```java
public class KafkaIO {

  public static class Read< K, V> extends PTransform<PBegin, PCollection<KV<K, V>>> {
    private Read(Coder<K> keyCoder, Coder<V> valueCoder, ...) {
    }

    public ReadWithMetadata<K, V> withMetadata() {
        return new ReadWithMetadata(...);
    }
  }

  public static class ReadWithMetadata< K, V> extends PTransform<PBegin, PCollection<KafkaRecord<K, V>>> {
    private ReadWithMetadata(Coder<K> keyCoder, Coder<V> valueCoder, ...) {...}

    public Read<K, V> withoutMetadata() {
        return new Read(...);
    }
  }
}
```

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaSource.java
```diff
@@ -0,0 +1,806 @@
...
+
+    /**
+     * Set a {@link ConsumerConfig} configuration property.
+     *
+     * @param configKey configuration property name
+     * @param configValue value for configuration property
+     * @return Builder
+     */
+    public Builder<K, V> withConsumerProperty(String configKey, Object configValue) {
+      Preconditions.checkArgument(!ignoredConsumerProperties.containsKey(configKey),
```
dhalperi: Following [the advice of the Guava team](https://github.com/google/guava/wiki/PreconditionsExplained), we usually import these statically.

```java

import static ...Preconditions.checkArgument;
...
checkArgument(!ignoredConsumerProperties.containsKey(configKey), "message");
```

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaSource.java
```diff
@@ -0,0 +1,806 @@
...
+          "No need to configure '%s'. %s", configKey, ignoredConsumerProperties.get(configKey));
+      mutableConsumerConfig.put(configKey, configValue);
+      return this;
```
dhalperi: we strongly discourage mutating these builders. Instead, it probably makes sense to just take a fully-formed `Map<String, Object>`.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaSource.java
```diff
@@ -0,0 +1,806 @@
...
+    }
+
+    /**
+     * Update consumer config properties. Note that this does not not discard already configured.
+     * Same as invoking #withConsumerProperty() with each entry.
```
dhalperi: `{@link #withConsumerProperty}` or something like that ?

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaSource.java
```diff
@@ -0,0 +1,806 @@
...
+     */
+    public Builder<K, V> withConsumerProperties(Map<String, Object> configUpdate) {
+      for (Entry<String, Object> e : configUpdate.entrySet()) {
+        withConsumerProperty(e.getKey(), e.getValue());
+      }
+      return this;
+    }
+
+    /**
+     * Set Coder for Key.
```
dhalperi: `{@link Coder}`

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaSource.java
```diff
@@ -0,0 +1,806 @@
...
+          "No need to configure '%s'. %s", configKey, ignoredConsumerProperties.get(configKey));
+      mutableConsumerConfig.put(configKey, configValue);
+      return this;
+    }
+
+    /**
+     * Update consumer config properties. Note that this does not not discard already configured.
+     * Same as invoking #withConsumerProperty() with each entry.
+     *
+     * @param configUpdate updates to {@link ConsumerConfig}
```
dhalperi: generally, feel free to drop `@param` and `@return` tags -- relatively unused in google style.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaSource.java
```diff
@@ -0,0 +1,806 @@
...
+     * @return Builder
+     */
+    public Builder<K, V> withValueCoder(Coder<V> valueCoder) {
+        this.valueCoder = valueCoder;
+        return this;
+    }
+
+    /**
+     * A function to assign a timestamp to a record. When this is not set, processing timestamp
+     * (when record is processed by {@link UnboundedReader#advance()}) is used.
```
dhalperi: better just "defaults to processing time".

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaSource.java
```diff
@@ -0,0 +1,806 @@
...
+    }
+
+    /**
+     * Build Unbounded Kafka Source
+     *
+     * @return UnboundedKafkaSource
+     */
+    public UnboundedSource<KafkaRecord<K, V>, KafkaCheckpointMark> build() {
+
+      ImmutableMap<String, Object> consumerConfig = ImmutableMap.copyOf(mutableConsumerConfig);
```
dhalperi: I don't believe this is deeply immutable, so may not actually be immutable.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaSource.java
```diff
@@ -0,0 +1,806 @@
...
+      this.topics = topics;
+      this.keyCoder = keyCoder;
+      this.valueCoder = valueCoder;
+      this.timestampFn = timestampFn;
+      this.watermarkFn = watermarkFn;
+      this.assignedPartitions = ImmutableList.copyOf(assignedPartitions);
+    }
+
+    @Override
+    public List<UnboundedKafkaSource<K, V>> generateInitialSplits(
```
dhalperi: probably worth javadoccing the splitting behavior extensively here, and briefly at the class level (with a `{@link }` pointer to here). Users will want to understand how the source is scalabale

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaSource.java
```diff
@@ -0,0 +1,806 @@
...
+        public int compare(TopicPartition tp1, TopicPartition tp2) {
+          return ComparisonChain
+              .start()
+              .compare(tp1.topic(), tp2.topic())
+              .compare(tp1.partition(), tp2.partition())
+              .result();
+        }
+      });
+
+      Preconditions.checkArgument(desiredNumSplits > 0);
```
dhalperi: message

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaSource.java
```diff
@@ -0,0 +1,806 @@
...
+      Preconditions.checkState(partitions.size() > 0,
+          "Could not find any partitions. Please check Kafka configuration and topic names");
+
+      int numSplits = Math.min(desiredNumSplits, partitions.size());
+
+      List<List<TopicPartition>> assignments = Lists.newArrayList();
```
dhalperi: just `new ArrayList<>();` -- the Guava team's recommendation is to stop using these static factory methods since Java 7 has the diamond (`<>`) syntax.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaSource.java
```diff
@@ -0,0 +1,806 @@
...
+
+          if (!curBatch.hasNext()) {
+            return false;
+          }
+        }
+      }
+    }
+
+    private static <T> T decode(byte[] bytes, Coder<T> coder) throws IOException {
+      if (bytes == null) {
```
dhalperi: can bytes be null? That seems strange. What would that mean?

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaSource.java
```diff
@@ -0,0 +1,806 @@
...
+        }
+      };
+    }
+  }
+
+  /**
+   * Usually the users are only interested in value in KafkaRecord. This is a convenient class
+   * to strip out other fields in KafkaRecord returned by UnboundedKafkaValueSource
+   */
+  private static class UnboundedKafkaValueSource<V>
```
dhalperi: Oh man. There has to be a way to make it possible that we only have a single implementation of the Source/Reader but we can support both with/without record metadata. I'll think and get back to you.

rangadi:
added unit tests (CountingSourceTest adapted to KafkaSource).

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaSource.java
rangadi: Yeah. Since user provides multiple functions, the type affects the builder/reader etc.. I didn't see a good way to avoid these wrappers.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaSource.java
rangadi: 'key' could be null. not sure about value (mostly not).

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaSource.java
rangadi: I see. will remove them. Diamond systax does not seem to be effect except for defintions. E.g. ```x.addList(new ArrayList<>())``` requires explicit type

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaSource.java
rangadi: yep. that is important. I will add.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaSource.java
rangadi: I didn't have those initially. ```mvn javadoc:javadoc``` warns without those.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaSource.java
rangadi: will look into this.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaSource.java
rangadi: these don't look as bad with Java 8.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaSource.java
rangadi: I agree, a bounded source would be quite useful. 
will also look into making it a PTransform.

rangadi:
multiple updates
- KafkaIO refactor : interface changes to return a transform (biggest part of change).
- allow specifying specific partitions (rather than just topics)
- use KV in KafkaRecord
- address most or all of review comments from Dan.

dhalperi:
@rangadi I know you've made some threading and performance improvements in the last day or so. Ping me here once the code is ready again.

contrib/examples/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/examples/TopHashtagsExample.java
```diff
@@ -0,0 +1,262 @@
...
+  private static final Logger LOG = LoggerFactory.getLogger(TopHashtagsExample.class);
+
+  /**
+   * Options for the app.
+   */
+  public static interface Options extends PipelineOptions {
+    @Description("Sliding window size, in minutes")
+    @Default.Integer(10)
+    Integer getSlidingWindowSize();
+    void setSlidingWindowSize(Integer value);
```
dhalperi: @kennknowles any suggestions on naming here? Maybe `--slidingWindowLengthMinutes` and `--slidingWindowIntervalMinutes`?

@rangadi in general, prefer units in variables when they are not super obvious. E.g., `timeoutMs` instead of just `timeout`

contrib/examples/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/examples/TopHashtagsExample.java
```diff
@@ -0,0 +1,262 @@
...
+
+    @Description("Trigger window period, in minutes")
+    @Default.Integer(1)
+    Integer getSlidingWindowPeriod();
+    void setSlidingWindowPeriod(Integer value);
+
+    @Description("Bootstarp Server(s) for Kafka")
```
dhalperi: typo bootstrap

contrib/examples/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/examples/TopHashtagsExample.java
```diff
@@ -0,0 +1,262 @@
...
+    @Required
+    String getBootstrapServers();
+    void setBootstrapServers(String servers);
+
+    @Description("One or more topics to read from")
```
dhalperi: ".. comma separated"

contrib/examples/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/examples/TopHashtagsExample.java
```diff
@@ -0,0 +1,262 @@
...
+  public static void main(String args[]) {
+
+    Options options = PipelineOptionsFactory.fromArgs(args).withValidation().as(Options.class);
+
+    Pipeline pipeline = Pipeline.create(options);
+
+    final int windowSize = options.getSlidingWindowSize();
+    final int windowPeriod = options.getSlidingWindowPeriod();
+
+    Read<?, String> reader = KafkaIO.read()
```
dhalperi: prefer `KafkaIO.Read` rather than static importing `Read`. Lots of things have a `Read` class so it makes the snippet clearer.

rangadi:
Updated the patch. 'Length Minutes' and 'Interval Minutes' sound better than 'Size' and 'Period' for window arguments.

contrib/examples/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/examples/TopHashtagsExample.java
```diff
@@ -0,0 +1,257 @@
...
+    @Default.Integer(1)
+    Integer getSlidingWindowIntervalMinutes();
+    void setSlidingWindowIntervalMinutes(Integer value);
+
+    @Description("Bootstrap Server(s) for Kafka")
+    @Required
+    String getBootstrapServers();
+    void setBootstrapServers(String servers);
+
+    @Description("One or more comma seperated topics to read from")
```
dhalperi: typo: separated.

contrib/examples/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/examples/TopHashtagsExample.java
```diff
@@ -0,0 +1,257 @@
...
+    Pipeline pipeline = Pipeline.create(options);
+
+    KafkaIO.Read<?, String> reader = KafkaIO.read()
+        .withBootstrapServers(options.getBootstrapServers())
+        .withTopics(options.getTopics())
+        .withValueCoder(StringUtf8Coder.of())
+        .withTimestampFn(timestampFn);
+
+    pipeline
+      .apply("sample_tweets", reader)
```
dhalperi: why only supply a name for this stage?

contrib/examples/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/examples/TopHashtagsExample.java
```diff
@@ -0,0 +1,257 @@
...
+                .withoutDefaults())
+      .apply(ParDo.of(new OutputFormatter()))
+      .apply(ParDo.of(new KafkaWriter(options)));
+
+    pipeline.run();
+  }
+
+  // The rest of the file implements DoFns to extract hashtags, formatting output, writing output
+  // back to Kafka. Note that writing to Kafka is not a complete Dataflow Sink. It is a best-effort
+  // logging of the results.
```
dhalperi: Take a grammar pass over this paragraph?

Drop the bit about the sink? That doc goes on KafkaIO.Write if anywhere.

contrib/examples/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/examples/TopHashtagsExample.java
```diff
@@ -0,0 +1,257 @@
...
+   * Emit hashtags in the tweet (if any).
+   */
+  private static class ExtractHashtagsFn extends DoFn<String, String> {
+
+    @Override
+    public void processElement(ProcessContext ctx) throws Exception {
+      for (JsonNode hashtag : jsonMapper.readTree(ctx.element())
+          .with("entities")
+          .withArray("hashtags")) {
+        ctx.output(hashtag.get("text").asText());
```
dhalperi: You might want to check your Eclipse configuration. It is making some wonky choices with respect to indentation etc. Obviously we're not going to enforce Google style in Beam, but we'd welcome it for contributions from Googlers like yourself ;)

Here's an open-source copy: https://github.com/google/styleguide/blob/gh-pages/eclipse-java-google-style.xml

contrib/examples/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/examples/TopHashtagsExample.java
```diff
@@ -0,0 +1,257 @@
...
+  /**
+   * Emit hashtags in the tweet (if any).
+   */
+  private static class ExtractHashtagsFn extends DoFn<String, String> {
+
+    @Override
+    public void processElement(ProcessContext ctx) throws Exception {
+      for (JsonNode hashtag : jsonMapper.readTree(ctx.element())
+          .with("entities")
+          .withArray("hashtags")) {
```
dhalperi: does Jackson gracefully handle the case when `entities` or `entities.hashtags` is missing, or will it crash? Do you want to catch and ignore?

contrib/examples/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/examples/TopHashtagsExample.java
```diff
@@ -0,0 +1,257 @@
...
+        ctx.output(hashtag.get("text").asText());
+      }
+    }
+  }
+
+  // extract timestamp from "timestamp_ms" field.
+  private static SerializableFunction<KV<byte[], String>, Instant> timestampFn =
```
dhalperi: `final`?

Maybe name it something like `TWEET_TIMESTAMP_OR_NOW`?

Maybe Javadoc it to explain what it does (timestamp or now)?

contrib/examples/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/examples/TopHashtagsExample.java
```diff
@@ -0,0 +1,257 @@
...
+            throw Throwables.propagate(e);
+          }
+        }
+      };
+
+  // return json string containing top hashtags and window information time
+  private static class OutputFormatter extends DoFn<List<KV<String, Long>>, String>
+      implements DoFn.RequiresWindowAccess {
+
+    private transient DateTimeFormatter formatter;
```
dhalperi: why not make this `private static final` rather than re-allocating per bundle? `DateTimeFormatter` is threadsafe, right?

contrib/examples/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/examples/TopHashtagsExample.java
```diff
@@ -0,0 +1,257 @@
...
+    private transient ObjectWriter jsonWriter;
```
dhalperi: same for `jsonWriter` -- also definitely threadsafe.

contrib/examples/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/examples/TopHashtagsExample.java
```diff
@@ -0,0 +1,257 @@
...
+      this.config = ImmutableMap.<String, Object>of(
+          "bootstrap.servers", options.getBootstrapServers(),
+          "key.serializer",    StringSerializer.class.getName(),
+          "value.serializer",  StringSerializer.class.getName());
+    }
+
+    @Override
+    public void processElement(ProcessContext ctx) throws Exception {
+      if (producer == null) {
+        producer = new KafkaProducer<String, String>(config);
```
dhalperi: Just allocate `producer` in `startBundle`, and keep the code simple.

contrib/examples/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/examples/TopHashtagsExample.java
rangadi: I had used this name with previous `Read.from(...).named()`. We could remove it.

contrib/examples/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/examples/TopHashtagsExample.java
```diff
@@ -0,0 +1,257 @@
...
+      }
+      LOG.info("Top Hashtags : " + ctx.element());
```
dhalperi: Do not log per element -- super slow!

Just drop I think.

(If you were going to log, it would be `LOG.info("Top hashtags: {}", ctx.element());` -- for performance when logging is disabled.

contrib/kafka/pom.xml
```diff
@@ -0,0 +1,176 @@
...
+      <distribution>repo</distribution>
+    </license>
+  </licenses>
+
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <google-cloud-dataflow-version>[1.2.0,2.0.0)</google-cloud-dataflow-version>
+    <hamcrest.version>1.3</hamcrest.version>
+    <junit.version>4.11</junit.version>
+    <slf4j.version>1.7.7</slf4j.version>
```
dhalperi: are these not inherited from the parent?

I'd imagine I'll drop all these when I import to Beam

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaCheckpointMark.java
```diff
@@ -0,0 +1,80 @@
...
+  @Override
+  public void finalizeCheckpoint() throws IOException {
+    /*
+     * nothing to do.
+     *
+     * we might want to support committing offset in Kafka, though it does not guarantee
+     * no-duplicates, it could support Dataflow restart better. Unlike an update of a dataflow job,
+     * a restart does not have checkpoint state. This secondary checkpoint might be a good start
+     * for readers. Another similar benefit is when the number of workers or number of
+     * Kafka partitions changes.
```
dhalperi: I would, via syntax, logically separate these two points.

```java
... {
    /* do nothing */

    // TODO: We might want to support committing the consumer offset to the Kafka broker. This would mean that
    // if we restarted the job fresh we would not get the old data.
```

I honestly did not actually understand your paragraph, so the above is my best guess. A few questions about what you wrote:

* would the restarted job use the same consumer groups?
* why not just let the user specific an optional offset for each topic/partition instead? That functionality seems like it would completely subsume any benefits of committing the consumer offset. Not that I think you should make this change -- I don't -- but I'm trying to understand the rationale for this TODO vs other possibilities.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaCheckpointMark.java
```diff
@@ -0,0 +1,80 @@
...
+     */
+  }
+
+  /**
+   * TopicPartition, offset tuple. Defines specific location in the partitions.
+   */
```
dhalperi: doesn't parse -- try with complete sentences?

I think this is just a bean-like object to hold the topic, partition, and offset that comprise the checkpoint mark for a single partition. Easy to write out as a whole sentence. :)

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaCheckpointMark.java
```diff
@@ -0,0 +1,80 @@
...
+import com.google.cloud.dataflow.sdk.io.UnboundedSource;
+
+import org.apache.kafka.common.TopicPartition;
+
+import java.io.IOException;
+import java.io.Serializable;
+import java.util.List;
+
+/**
+ * Checkpoint for an unbounded KafkaSource reader. Consists of Kafka topic name, partition id,
```
dhalperi: Checkpoint for an unbounded `KafkaIO.Read`.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
```diff
@@ -0,0 +1,928 @@
...
+import java.util.Map;
+import java.util.NoSuchElementException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.SynchronousQueue;
+import java.util.concurrent.TimeUnit;
+
+import javax.annotation.Nullable;
+
+/**
```
dhalperi: please flesh out. Use https://github.com/apache/incubator-beam/blob/master/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/bigtable/BigtableIO.java#L66 as a reference.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
```diff
@@ -0,0 +1,928 @@
...
+ * Dataflow support for consuming Kafka topics as {@link Unbounded} sources.
+ *
+ * <p>
+ * The Kafka partitions
+ * are evenly distributed among splits (workers). Dataflow checkpointing is fully supported and
+ * each split can resume from previous checkpoint. See
+ * {@link UnboundedKafkaSource#generateInitialSplits(int, PipelineOptions)} for more details on
+ * splits and checkpoint support.
+ *
+ * <pre>{@code
```
dhalperi:
```
<h3>Reading from Kafka</h3>

To read from a Kafka broker, you must configure the bootstrap servers and one or more Kafka topics to consume.

<pre>{@code
Pipeline p = ...;
PCollection<byte[], byte[]> unboundedMessages =
    p.apply("Read two topic_a and topic_b",
        KafkaIO.read().withBootstrapServers("..").withTopics(...));
...
}</pre>

<p>You may optionally supply a `Coder<V>` value coder to determine the decoding of the message:

<pre>{@code
PCollection<byte[], V> messages = p.apply...
}</pre>
...

<p>To produce a bounded {@code PCollection}, use {@link KafkaIO.Read#withMaxNumRecords} or
...:

<pre>{@code
...
}</pre>

```

etc.

A few guidelines:

* Spell out the basic use cases in logical separation (source before sink, unbounded before bounded) and in ascending order of complexity (default configuration, overriding the value coder, overriding the key coder, supplying custom timestamps). Do not do it all at once.

* Try to put the minimal content that is both necessary (include that `p` is a `Pipeline`) and sufficient (drop the `Values.<String>create()`).

Any gotchas you can think of, document them here.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
```diff
@@ -0,0 +1,928 @@
...
+          }
+        };
+
+    private Read(
+        List<String> topics,
+        List<TopicPartition> topicPartitions,
+        Coder<K> keyCoder,
+        Coder<V> valueCoder,
+        @Nullable SerializableFunction<KV<K, V>, Instant> timestampFn,
+        @Nullable SerializableFunction<KV<K, V>, Instant> watermarkFn,
```
dhalperi: I would not make these `null` -- I would instead put the defaults in the `KafkaIO.read()` static factory method.

I would in fact enforce that they are not `null`.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
```diff
@@ -0,0 +1,928 @@
...
+      checkNotNull(watermarkFn);
+      return new Read<K, V>(topics, topicPartitions, keyCoder, valueCoder,
+          timestampFn, watermarkFn, consumerFactoryFn, consumerConfig, maxNumRecords, maxReadTime);
+    }
+
+    /**
+     * A factory to create Kafka {@link Consumer} from consumer configuration.
+     * Mainly used for tests. Default is {@link KafkaConsumer}.
+     */
+    public Read<K, V> withConsumerFactoryFn(
```
dhalperi: if it's only used for tests, make it package-private and annotate with `@VisibleForTesting`.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
```diff
@@ -0,0 +1,928 @@
...
+            ConsumerConfig.RECEIVE_BUFFER_CONFIG, 512 * 1024,
+
+            // default to latest offset when we are not resuming.
+            ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "latest",
+            // disable auto commit of offsets. we don't require group_id. could be enabled by user.
+            ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG, false);
+
+    // default Kafka 0.9 Consumer supplier. static variable to avoid capturing 'this'
+    private static SerializableFunction<Map<String, Object>, Consumer<byte[], byte[]>>
+      kafka9ConsumerFactory =
```
dhalperi: Why isn't this `private static final KAFKA_9_CONSUMER_FACTORY_FN`?

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
```diff
@@ -0,0 +1,928 @@
...
+    private final List<TopicPartition> topicPartitions; // mutually exclusive with topics
+    private final Coder<K> keyCoder;
+    private final Coder<V> valueCoder;
+    @Nullable private final SerializableFunction<KV<K, V>, Instant> timestampFn;
+    @Nullable private final SerializableFunction<KV<K, V>, Instant> watermarkFn;
+    private final
+      SerializableFunction<Map<String, Object>, Consumer<byte[], byte[]>> consumerFactoryFn;
+    private final Map<String, Object> consumerConfig;
+    private final long maxNumRecords; // bounded read, mainly for testing
+    private final Duration maxReadTime; // bounded read, mainly for testing
```
dhalperi: drop the "mainly for testing" part -- it may be true as far as we know, but we probably don't want to put value judgements on the user's pipeline decisions.

(Btw, I acknowledge that I think I've written comments that violate what I just wrote in the past. With time comes wisdom.... or so I hope)

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
```diff
@@ -0,0 +1,928 @@
...
+     */
+    private static final Map<String, String> ignoredConsumerProperties = ImmutableMap.of(
+        ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, "Set keyDecoderFn instead",
+        ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, "Set valueDecoderFn instead"
+        // "group.id", "enable.auto.commit", "auto.commit.interval.ms" :
+        //     lets allow these, applications can have better resume point for restarts.
+        );
+
+    // set config defaults
+    private static final Map<String, Object> defaultConsumerProperties =
```
dhalperi: DEFAULT_CONSUMER_PROPERTIES

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
```diff
@@ -0,0 +1,928 @@
...
+    private final
+      SerializableFunction<Map<String, Object>, Consumer<byte[], byte[]>> consumerFactoryFn;
+    private final Map<String, Object> consumerConfig;
+    private final long maxNumRecords; // bounded read, mainly for testing
+    private final Duration maxReadTime; // bounded read, mainly for testing
+
+    /**
+     * A set of properties that are not required or don't make sense for our consumer.
+     */
+    private static final Map<String, String> ignoredConsumerProperties = ImmutableMap.of(
```
dhalperi: IGNORED_CONSUMER_PROPERTIES

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
dhalperi: https://google.github.io/styleguide/javaguide.html#s5.2.4-constant-names

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
```diff
@@ -0,0 +1,928 @@
...
+
+    /**
+     * Returns a new {@link Read} with {@link Coder} for key bytes.
+     * <p> Since this changes the type for key, settings that depend on the type
+     * ({@link #withTimestampFn(SerializableFunction)} and
+     *  {@link #withWatermarkFn(SerializableFunction)}) make sense only after the coders are set.
+     */
+    public <KeyT> Read<KeyT, V> withKeyCoder(Coder<KeyT> keyCoder) {
+      checkState(timestampFn == null, "Set timestampFn after setting key and value coders");
+      checkState(watermarkFn == null, "Set watermarkFn after setting key and value coders");
```
dhalperi: okay, I see this may invalidate some earlier comments of mine. this is definitely an interesting and thorny issue...

let's sync tomororw, but you may well have found a perfectly reasonable solution to this :)

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
```diff
@@ -0,0 +1,928 @@
...
+
+  private static class IdentityFn<T> implements SerializableFunction<T, T> {
+    @Override
+    public T apply(T input) {
+      return input;
+    }
+  }
+
+  /**
+   * Creates and uninitialized {@link Read} {@link PTransform}. Before use, basic Kafka
```
dhalperi: an

rangadi:
Wow! thanks for such a detailed review. I will address it.

Added support for reporting backlog. Since I added one thread for kafka I/O anyway, added one more for fetching offsets.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
dhalperi: note that I now reject this, given the issues with changing different typed parameters at the same time.

contrib/examples/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/examples/TopHashtagsExample.java
rangadi: rephrased and removed reference to sink.

contrib/examples/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/examples/TopHashtagsExample.java
rangadi: done.

contrib/examples/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/examples/TopHashtagsExample.java
rangadi: done. that is better.

contrib/examples/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/examples/TopHashtagsExample.java
rangadi: done. and should I close the producer in finishBundle()?

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaCheckpointMark.java
rangadi: That is correct. I rewrote the comment close to what you suggested. 
 - The restarted job need to use the same consumer group ("client_id")
 - yes, this is effectively same as letting users specifying an offset for a partition. 

The third case is handling the case when the number of partitions or Dataflow splits changes across an update. The committed offsets will let us handle changes in distribution of partitions among splits. Thats what the latter part of the my comment was about.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaCheckpointMark.java
rangadi: Yep, that is better.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
rangadi: Yes, BigtableIO JavaDoc is so much better. I need spend more time on this. will get back to it.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
rangadi: Thanks! this will be very useful. will do.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
rangadi: This is required to support another version of the consumer (e.g. Kafka 0.8.x Consumer).

rangadi:
@dhalperi Updated Read interface as we discussed. Needed changes to tests as well.
TODO: good JavaDoc for KafkaIO as you suggested. will update later today.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/bigtable/BigtableIO.java
```diff
@@ -72,7 +72,7 @@
  * <h3>Reading from Cloud Bigtable</h3>
  *
  * <p>The Bigtable source returns a set of rows from a single table, returning a
- * {@code PCollection&lt;Row&gt;}.
```
dhalperi: please revert

contrib/examples/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/examples/TopHashtagsExample.java
```diff
@@ -0,0 +1,265 @@
...
+    @Override
+    public void startBundle(Context c) throws Exception {
+      if (producer == null) { // in Beam, startBundle might be called multiple times.
+        producer = new KafkaProducer<String, String>(config);
+      }
+    }
+
+    @Override
+    public void finishBundle(Context c) throws Exception {
+      producer.close();
```
dhalperi: If the producer is closed in `finishBundle` will it need to be reopened in `startBundle`? Is there a sort of `flush()` instead?

contrib/examples/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/examples/TopHashtagsExample.java
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    @Override
+    public void processElement(ProcessContext ctx) throws Exception {
+      LOG.info("Top Hashtags : {}", ctx.element());
```
dhalperi: Logging per element is considered harmful! Remove or downgrade to trace?

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
```diff
@@ -0,0 +1,1046 @@
...
+      }
+
+      return result;
+    }
+
+    @Override
+    public UnboundedKafkaReader<K, V> createReader(PipelineOptions options,
+                                                   KafkaCheckpointMark checkpointMark) {
+      if (assignedPartitions.isEmpty()) {
+        LOG.warn("hack: working around DirectRunner issue. It does not generateSplits()");
```
dhalperi: @tgroh opinions here?

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
```diff
@@ -0,0 +1,1046 @@
...
+    }
+
+    @Override
+    public Coder<KafkaCheckpointMark> getCheckpointMarkCoder() {
+      return SerializableCoder.of(KafkaCheckpointMark.class);
+    }
+
+    @Override
+    public boolean requiresDeduping() {
+      return false;
```
dhalperi: should comment why this is false.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
```diff
@@ -0,0 +1,1046 @@
...
+        checkState(checkpointMark.getPartitions().size() == source.assignedPartitions.size(),
+            "checkPointMark and assignedPartitions should match");
+        // we could consider allowing a mismatch, though it is not expected in current Dataflow
+
+        for (int i = 0; i < source.assignedPartitions.size(); i++) {
+          PartitionMark ckptMark = checkpointMark.getPartitions().get(i);
+          TopicPartition assigned = source.assignedPartitions.get(i);
+
+          checkState(ckptMark.getTopicPartition().equals(assigned),
+              "checkpointed partition %s and assinged partition %s don't match at position %d",
```
dhalperi: typo: assigned

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
```diff
@@ -0,0 +1,1046 @@
...
+          partitionStates.get(i).consumedOffset = ckptMark.getOffset();
+        }
+      }
+    }
+
+    private void consumerPollLoop() {
+      // Read in a loop and enqueue the batch of records, if any, to availableRecordsQueue
+      while (!closed) {
+        try {
+          ConsumerRecords<byte[], byte[]> records = consumer.poll(1000);
```
dhalperi: extract 1000 as constant and comment

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
```diff
@@ -0,0 +1,1046 @@
...
+          if (!records.isEmpty()) {
+            availableRecordsQueue.put(records); // blocks until dequeued.
+          }
+        } catch (InterruptedException e) {
+          LOG.warn(this + " consumer thread is interrupted", e); // not expected
```
dhalperi: use `LOG.warn("{} consumer thread", this, e);` instead

dhalperi: or better yet make all the logs in this file have the same scheme. Below it's `Log.WARN("{}: ...");`

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
```diff
@@ -0,0 +1,1046 @@
...
+
+      LOG.info("{} : Returning from consumer pool loop", this);
+    }
+
+    private void nextBatch() {
+      curBatch = Collections.emptyIterator();
+
+      ConsumerRecords<byte[], byte[]> records;
+      try {
+        records = availableRecordsQueue.poll(10, TimeUnit.MILLISECONDS);
```
dhalperi: constant

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
```diff
@@ -0,0 +1,1046 @@
...
+          }
+
+          ConsumerRecord<byte[], byte[]> rawRecord = pState.recordIter.next();
+          long consumed = pState.consumedOffset;
+          long offset = rawRecord.offset();
+
+          if (consumed >= 0 && offset <= consumed) { // -- (a)
+            // this can happen when compression is enabled in Kafka (seems to be fixed in 0.10)
+            // should we check if the offset is way off from consumedOffset (say > 1M)?
+            LOG.warn("ignoring already consumed offset {} for {}", offset, pState.topicPartition);
```
dhalperi: this is fairly expected, right? Should we disable these warnings so that we don't log per element?

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
```diff
@@ -0,0 +1,1046 @@
...
+            }
+          }, 0, OFFSET_UPDATE_INTERVAL_SECONDS, TimeUnit.SECONDS);
+
+      return advance();
+    }
+
+    @Override
+    public boolean advance() throws IOException {
+      /* Read first record (if any). we need to loop here because :
+       *  - (a) some records initially need to be skipped if they are before consumedOffset
```
dhalperi: is it possible to move all the skipping logic into `start()`? Then `advance()` would be much cleaner I think.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
```diff
@@ -0,0 +1,1046 @@
...
+    // update latest offset for each partition.
+    // called from offsetFetcher thread
+    private void updateLatestOffsets() {
+      for (PartitionState p : partitionStates) {
+        try {
+          offsetConsumer.seekToEnd(p.topicPartition);
+          long offset = offsetConsumer.position(p.topicPartition);
+          p.setLatestOffset(offset);;
+        } catch (Exception e) {
+          LOG.warn(this + " : exception while fetching latest offsets. ignored.",  e);
```
dhalperi: don't construct strings with '+' in log statements, see other similar content. Please skim entire PR.

dhalperi:
R: @dpmills @mshields822 

Daniel has already taken a look at this a month back, can you re-skim it?

Mark has been implementing PubSub. Mark, can you take a look at the Kafka source and see what lessons you can impart?

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
```diff
@@ -0,0 +1,1046 @@
...
+ * the License.
+ */
+
+package com.google.cloud.dataflow.contrib.kafka;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
+import static com.google.common.base.Preconditions.checkState;
+
+import com.google.api.client.repackaged.com.google.common.annotations.VisibleForTesting;
```
mshields822: No need for repackaged.

mshields822:
LGTM
No overlap with pub/sub impl:
 - Kafka: Due to strict partitioning cannot use additional source splits to naively hide get latency. So need background thread. Pubsub: No background threads needed.
 - Kafka: Since seekable (modulo caveats) can estimate backlog. Pubsub: Not  seekable and pub/sub don't publish any backlog estimate (at least publically). So backlog is just what's received but not yet read by advance in the reader.
 - Kafka: Imposes key/value structure on records. Pub/sub: uninterpreted bytes.
 - Kafka: Order preserving, so assuming (implied) timestamps are monotonic watermark is just current records timestamp. Pub/sub: Not order preserving, no assumption on timestamp monotonicity, need to estimate watermark.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
tgroh: This message shouldn't be at `WARN`; potentially keep, but change the message to note that `generateInitialSplits` wasn't called, which will cause (**BAD THINGS**) to occur, or move to a lower log level

As there's no requirement to actually generate splits, this is instead a fallback to initialize in the case where we call the initial source. Given that it appears that you need the Consumer, which may require information only available at execution time, this seems like a reasonable workaround for that - if it's possible to generate some shorthand `TopicPartition` for each topic that is "100% of this topic", it would be preferable to construct that at source construction time and put it in the `assignedPartitions`, and then remove this block

In the future, we plan on exercising `generateInitialSplits` in the `InProcessPipelineRunner`, but may skip it if the number of target splits is equal to 1, so this would likely have to stick around to handle that case.

rangadi: `generateInitialSplits()` is public API and it is better if the contract is clear about when it is called. [ My personal preference is not to skip it in DirectRunning  (it can run multiple splits in sequence if more than one is returned), but that is a separate issue, what if I have a source that does not want to have less than 3 splits?]

contrib/examples/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/examples/TopHashtagsExample.java
rangadi: There is a `flush()`. But we would still end up with many unclosed producers. How about making it static?

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
rangadi: Updated JavaDoc with code sample with examples.

contrib/examples/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/examples/TopHashtagsExample.java
rangadi: done. I logged it since it is only once every minute by default.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
rangadi: Done. Updated the all the LOG lines in the reader. 
JavaDoc was not clear about how exception is rendered with multiple objects as in `LOG.warn("msg", o1, exception)`. Just tried it, it does the right thing.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
dhalperi: A little more context:

This is not just the `DirectPipelineRunner` -- any runner may choose not to call `generateInitialSplits`, and the source needs to support reading in that case.

I might follow Thomas' suggestion and go with something like

```java
LOG.warn("It looks like the runner did not call generateInitialSplits on this {}. This means that the Kafka stream will not be consumed in parallel.", this);
```

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
dhalperi: http://stackoverflow.com/a/6374166/1715495

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
rangadi: Yep. I was going to update the message.
still I am not a fan of selectively not calling `generateInitialSplits()`. I think is better for the runner to figure out how to handle multiple splits than requiring source developers to special case it. Related issue : a source can return more splits in `desiredSplits`.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
rangadi: Thanks. Confirmed the behavior.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
rangadi: It is fixed in 10 and may be very near 0.9.x. It was essentially a bug in the consumer. It has enough information to skip these internally. 

I wish there was a way to expose such metrics through stats. I think we can leave it for now. One degenerate case (which I did hit) is when I reset kafka cluster while the dataflow job is running. Since the offset is reset back to 0, all the records were before the recently read offset

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
rangadi: We would have to loop in both cases. There is guarantee start() would be able to read all of them without blocking.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
rangadi: fixed.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
rangadi: oops! Thanks for catching.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
rangadi: done

rangadi:
Thanks @mshields822. 
One thing I would like to expand is that Kafka wouldn't be much different w.r.t timestamp and wantermarks. The order is only preserved within a partition, and we don't guarantee synchronous progress across multiple partitions.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
dhalperi: needs work

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
dhalperi: needs work.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
dhalperi: needs work

dhalperi:
Looks like you need to push? I'm going to stop commenting on outstanding feedback :)

rangadi:
sorry, forgot to push before heading to lunch. Addressed all the comments except last one. I will double check any remaining comments.

contrib/kafka/pom.xml
rangadi: This is not addressed yet. I had some issue trying to use parent pom. I will check.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/bigtable/BigtableIO.java
rangadi: done. Eclipse didn't render this properly.

contrib/kafka/src/main/java/com/google/cloud/dataflow/contrib/kafka/KafkaIO.java
```diff
@@ -0,0 +1,1053 @@
...
+          PartitionState pState = curBatch.next();
+
+          if (!pState.recordIter.hasNext()) { // -- (c)
+            pState.recordIter = Collections.emptyIterator(); // drop ref
+            curBatch.remove();
+            continue;
+          }
+
+          ConsumerRecord<byte[], byte[]> rawRecord = pState.recordIter.next();
+          long consumed = pState.consumedOffset;
```
dhalperi: Is there a bound on the number of records that we'll have to skip? We can only log a few 1000 per second, so if it's millions then logging each record will induce large delays in starting up.

Alternative would be to set a flag and only log once.

rangadi: It would fairly small (10s to 100s). The upper limit is the number of kafka records compressed together. Future versions of KafkaConsumer already fixed this issue with compressed messages and they skip them. So it is not at all expected in near future.

One case where it can cause millions of records is the case I mention above where Kafka is restarted from scratch which resets the offsets while the Dataflow app is running. Not sure weather we want to handle that case. It is better for the user to realize this problem (should be very rare), and take appropriate action (may be just restarting the Dataflow app).

That said I can certainly add a flag to limit it to one message.

dhalperi:
Closing in favor of https://github.com/apache/incubator-beam/issues/142

PR #124, opened 2016-03-02
Move Java 8 tests to their own module

kennknowles:
This allows easy setting of the compiler source and target version to
1.8 without any fragile plugin configuration. It also allows the
import into Eclipse to easily set separate compliance levels, as this
is done per-Eclipse-project.


kennknowles:
Going to wait to submit to beam an backport.

PR #126, opened 2016-03-04
Remove hardcoded package name in CoderUtils

kennknowles:


kennknowles:
Going to wait to submit to beam an backport.

PR #145, opened 2016-03-11
Cassandra source main and  test classes

TechM-Google:
CassandraIO and CassandraReadIO  main and test classes


TechM-Google:
A new pull requested with updated files is created

PR #149, opened 2016-03-16
Add custom Kinesis source to contrib

rtshadow:
Custom Kinesis source that lets you use Kinesis stream as a source for Dataflow pipeline. 
It will:

- parallellize reading from many shards into many UnboundedReader(s)
- give you proper checkpointing scheme in case of failures
- deaggregate records produced by Kinesis Producer Library

The main yet unimplemented feature is to handle shard split / merge events, that can occur on Kinesis stream.

Why to contrib instead of main source?

- this is just a custom Kinesis source. I think we cannot say we support Kinesis until there will be a custom Kinesis sink too.
- currently Kinesis source does not support split / merge events of the shard. I think this is also a crucial thing to have
- I've got bunch of integration and E2E tests, which I didn't know where to put in the main source (are there any such tests?)


rtshadow:
Hey,
It's been two weeks now since I've opened this pull request. Is there any progress on it?

Cheers,
Przemek

dhalperi:
Hi @rtshadow,

Sorry for the delay. Actually, this repository is not accepting new contribs now, as we move forward with Apache Beam. All development for new features has moved there. Hopefully, it should be easy to rebase on top of that repository and issue a pull request there. (apache/incubator-beam).

Contributors guide: http://beam.incubator.apache.org/contribution-guide/

Thanks!

PR #167, opened 2016-03-24
Backport apache/incubator-beam#65

dhalperi:
This depends on #166

PR #172, opened 2016-03-29
Validate that Pubsub resource names are at least 3 chars

dpmills:


dpmills:
r=@dhalperi for code review

dhalperi:
@davorbonaci yes I realized this means I have to rebase backport65 / #167 . It won't conflict, and I can handle it. But I will probably not do another backport until after we get the release rolled out.

PR #173, opened 2016-03-29
Add ModelEnforcements

tgroh:
ModelEnforcements ensure that user-written code conforms to the model,
in order to ensure that it is portable between runners.

Add EncodabilityEnforcement and ImmutabilityEnforcement.
EncodabilityEnforcement ensures that all values can be encoded.
ImmutabilityEnforcement ensures that values are not mutated within the
processing of a Bundle.

Backports https://github.com/apache/incubator-beam/pull/74


kennknowles:
LGTM

PR #194, opened 2016-04-08
Move tempLocation to PipelineOptions.

peihe:
It is back ported from:
https://github.com/apache/incubator-beam/pull/44


peihe:
R: @dhalperi

dhalperi:
R: @davorbonaci 

Davor explicitly decided not to backport this in the Great Backporting of 2016 (tm) because it is not backwards-compatible.

davorbonaci:
It is backward-compatible. Merging.

PR #200, opened 2016-04-12
Implement BigQueryIO.Write as a sink

peihe:
This is back ported from:
https://github.com/apache/incubator-beam/pull/48

### DIFF
--- a/sdk/src/test/java/com/google/cloud/dataflow/sdk/io/BigQueryIOTest.java
+++ b/sdk/src/test/java/com/google/cloud/dataflow/sdk/io/BigQueryIOTest.java
@@ -20,6 +20,8 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNull;
 
 import com.google.api.client.util.Data;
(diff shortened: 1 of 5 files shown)


peihe:
It is not ready for review yet.
Need an override in dataflow runner.

sdk/src/test/java/com/google/cloud/dataflow/sdk/io/BigQueryIOTest.java
```diff
@@ -20,6 +20,8 @@
 import static org.junit.Assert.assertNull;
 
 import com.google.api.client.util.Data;
+import com.google.api.services.bigquery.model.TableFieldSchema;
+import com.google.api.services.bigquery.model.JobConfigurationLoad;
```
dhalperi: Looks like there was an error in backport that reordered imports. Please fix -- we'd like to run a sync today and this is blocker.

peihe: done 
PTAL

PR #203, opened 2016-04-15
[BEAM-188] Write: apply GlobalWindows first

dhalperi:
And do not supply a timestamp when outputting.

Note that this is safe because the functions in the Writer cannot access the window
or timestamp. When we add per-Window or similar functions to the sinks, we will
likely do so at a higher level.

This is a backport of apache/incubator-beam#188


dhalperi:
R: @bjchambers ; Note this is just a backport review.

PR #210, opened 2016-04-19
Remove InProcessBundle

tgroh:
InProcessBundle is an implementation detail of InProcessBundleFactory,
which it should be contained within.

Backports [BEAM #203](https://github.com/apache/incubator-beam/pull/203)


tgroh:
R: @bjchambers

PR #206, opened 2016-04-15
Fix BigQuery.Write tempFilePrefix concatenation

peihe:

PR #211, opened 2016-04-19
Add WindowedValue#explodeWindows

tgroh:
This takes an existing WindowedValue and returns a Collection of
WindowedValues, each of which is in exactly one window.

Use the explode implementation on DoFnRunnerBase

Backports [BEAM #206](https://github.com/apache/incubator-beam/pull/206)


tgroh:
R: @bjchambers

PR #212, opened 2016-04-21
[BEAM-204] Backport https://github.com/apache/incubator-beam/pull/221/files

mshields822:
R: @davorbonaci 

Should consider for 1.5.2 if there is one.


dhalperi:
Thanks.

PR #214, opened 2016-04-21
Track pending elements via exploded WindowedValues

tgroh:
This allows the WindowedValues to be partially completed while the
still holding the watermark.

This backports [BEAM #207](https://github.com/apache/incubator-beam/pull/207)


tgroh:
R: @bjchambers

PR #219, opened 2016-04-26
Use java.util.Objects rather than com.google.common.repackaged

swegner:


swegner:
R: @dhalperi

dhalperi:
(Process note:) This did not need backporting from Beam because the corresponding Beam uses AutoValue and did not need MoreObjects.toStringHelper or Objects.equals

Issue #220, opened 2016-04-26
Cannot create Pubsub subscription when using DirectPipelineRunner and multiple projects.


rculbertson:
When using the DirectPipelineRunner, PubsubIO creates the subscription in the same project as the topic you want to read from. This only works if you have permissions to create subscriptions in that project. If you only have subscriber permissions for that topic, it fails with 403 `User not authorized to perform this action`. 

This works when using the DataflowPipelineRunner, because it creates the subscription in the project specified in DataflowPipelineOptions. This allows you to create the subscription in a project you own, and consume from a topic in another project where you have limited permissions.

davorbonaci:
Acknowledged. Thanks Ryan!

@tgroh, how complex would be to fix this?

dhalperi:
Hi Ryan, I believe this is fixed now (for a while..). Can you confirm?

PR #221, opened 2016-04-26
Create pubsub subscriptions in the project specified in dataflow options

rculbertson:
PubsubIO would create subscriptions in the same project as the topic. This fails for users who have permissions to subscribe to that topic, but do not have permissions to create subscriptions in that project. Instead create the subscription in the project specified in dataflow options. If no project is specified, then fallback to the topic project.

This is to address #220.

### DIFF
--- a/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/PubsubIO.java
+++ b/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/PubsubIO.java
@@ -50,6 +50,7 @@ import com.google.cloud.dataflow.sdk.values.PCollection.IsBounded;
 import com.google.cloud.dataflow.sdk.values.PDone;
 import com.google.cloud.dataflow.sdk.values.PInput;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableMap;
 
 import org.joda.time.Duration;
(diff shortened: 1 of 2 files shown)


rculbertson:
As requested, I signed the CLA.

dhalperi:
R: @mshields822 @dpmills 

Mark, can you look at this w.r.t. your Beam work replacing PubSubIO?

mshields822:
LGTM
There are a few other places where we have assumes subscription project = topic project. I'll try to track those down.

rculbertson:
@dhalperi Is this PR good to merge? It would be great to get this in the next release so we can remove the workarounds we currently have in place.

dpmills:
LGTM

This looks fine to merge.  One note, however: it won't fix streaming pipelines reading from Pubsub, since those don't use this code.  This is just a default implementation that the batch and direct runners use.  Fixing this for streaming pipelines will require changes to the Dataflow backend.

davorbonaci:
@dpmills, this should currently be working as expected on the `DataflowPipelineRunner`, so no changes should be required. Am I missing something?

@rculbertson, would you be willing to add a unit test to make sure this scenario doesn't regress?

The change itself is fine. Two things worth noting:
* We should forward port this change to Apache Beam.
* This introduces a dependency of PubsubIO (core SDK) on `DataflowPipelineOptions` (Dataflow runner). We are separately trying to untangle these types of dependencies. We can figure out separately what's the right way of solving this.

We'll try to get this integrated early next week, and make sure it is part of the next release (1.6.0).

rculbertson:
@davorbonaci Sure I can add a test sometime this week. Thanks!

rculbertson:
@davorbonaci I've added a test to `PubsubIOTest.java`

rculbertson:
@dhalperi @davorbonaci: Any update on getting this PR merged? Thanks!

rculbertson:
Looks like this functionality was added in #278. Closing PR.

dhalperi:
Ryan,

Thanks for the update. Is the new test you added still useful and relevant? If so, please rebase and reopen and I'll take a look!

Thanks!
Dan

PR #224, opened 2016-04-26
checkstyle.xml: blacklist repackaged code from Google API Client

dhalperi:
backported from apache/incubator-beam#245


bjchambers:
Is it worth also blacklisting org.apache.beam for the time being to prevent mishaps? What about "com.google.cloud.dataflow.repackaged"?

dhalperi:
both are plausibly good, in a separate PR to assume a clean import.

it would be insane if somehow we got dataflow.repackaged as an import in Dataflow code.

PR #227, opened 2016-04-27
Add withElements to CommittedBundle

tgroh:
When a bundle is partially completed, the unprocessed elements must be
placed in a new bundle to be processed at a later time. The bundle in
which they are processed should also have identical properties to the
bundle which the elements were initially present in.

withElements provides a simple way to create a "copy" of a bundle that
contains different elements.

This backports [BEAM #231](https://github.com/apache/incubator-beam/pull/231)


tgroh:
R: @bjchambers

dhalperi:
R: @bjchambers

PR #201, opened 2016-04-12
Add InProcessRegistrar

tgroh:
Adds @AutoService for the InProcessPipelineRunner and
InProcessPipelineOptions

This backports [BEAM #169](https://github.com/apache/incubator-beam/pull/169)

PR #228, opened 2016-04-27
Remove isKeyed property of InProcess Bundles

tgroh:
The property of keyedness belongs to a PCollection. A BundleFactory
propogates the key as far as possible, but does not track if a bundle is
keyed.

This backports [BEAM #201](https://github.com/apache/incubator-beam/pull/201)


tgroh:
R: @kennknowles 

I seem to have missed this one a while back

PR #229, opened 2016-04-27
Allow InProcess Evaluators to check Side Input completion

tgroh:
This checks to ensure that the PCollectionView in the SideInputWindow
for the provided window either has elements available or is empty.

Schedule a future to ensure that the SideInputWindows are appropriately
filled with an empty iterable after retreiving the element.

Backports [BEAM #220](https://github.com/apache/incubator-beam/pull/220)


tgroh:
R: @kennknowles for backport

PR #232, opened 2016-05-02
DataflowJavaSDK: remove all uses of Throwables.propagate

dhalperi:
It is slated to be deprecated:

https://github.com/google/guava/commit/49e7d716f0f72d00c490864c3d11b815cfe93ec4


davorbonaci:
LGTM

PR #234, opened 2016-05-02
Fix update sequence in InMemoryWatermarkManager

tgroh:
Because the WatermarkManager is not synchronized within the call to
updatePending, the sequence in which pending queues are updated must be
in such a manner as to add additional restrictions, then remove any
restrictions which no longer apply. This ensures any intermediate read
will see at worst a more restricted watermark than the actual watermark.

Backports [BEAM #264](https://github.com/apache/incubator-beam/pull/264)


tgroh:
R: @bjchambers

PR #235, opened 2016-05-02
Stop cloning coders in the InProcessRunner

tgroh:
This is excessively slow and also not useful, as coders are required to
be thread-safe

Backports [Beam #261](https://github.com/apache/incubator-beam/pull/261)


dhalperi:
R: @kennknowles

PR #237, opened 2016-05-03
Add CommittedResult

tgroh:
Return as the output to InProcessEvaluationContext#handleResult(). This
allows a richer return type to improve possible behaviors when a result
is returned.

Use this object in the InMemoryWatermarkManager.

Backports [Beam #260](https://github.com/apache/incubator-beam/pull/260)
Backports [Beam #265](https://github.com/apache/incubator-beam/pull/265)


tgroh:
R: @kennknowles

PR #238, opened 2016-05-03
Write: add limited logging

dhalperi:
This will help, for all sinks, users and developers gain insight into where time
is spent. (Enabling DEBUG level will provide more insight.)

This is a backport of apache/incubator-beam#279


dhalperi:
R: @lukecwik

PR #240, opened 2016-05-05
Firebase contrib: update PR and fix issues

dhalperi:
Hi @elibixby --

I've attempted to merge your PR but ran into some issues. I fixed several of them that were easy to handle, but I have a failing test (when run with `mvn verify` in the `firebase` directory).

These tests are very verbose, I've put the log in a gist: https://gist.github.com/dhalperi/8276fde921930323d80f38322be83f76


contrib/firebaseio/src/main/java/contrib/JacksonCoder.java
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  @Override
+  public List<? extends Coder<?>> getCoderArguments() {
+    return Collections.emptyList();
+  }
+
+  @Override
+  public void verifyDeterministic() {
+    return;
```
dhalperi: reminder that this is not deterministic if, e.g., there is a Java `Set` in the transform

elibixby:
I've seen that test failure before. It looks like the test is improperly recording the "reset" of the firebase as part of the output pipeline. I thought I had fixed this issue, since I've seen it before, once I have time I'll see if I can't reproduce and fix. I'm prepping for a conference so likely won't be able to get to it for a couple weeks.

elibixby:
Returning to this, also @sammcveety for FYI. I assume since you've resubmitted the PR, that this is still the correct way to submit code even post Apache Beam. I'd like to get this merged before I head off to PyCon since it sounds somewhat urgent. To whit, My AIs for this weekend to get this read to merge are:

* fixing the test above, 
* move from `Map.Entry<K, V>` to `KV<K, V>` types for the DoSet functions
* move from `DoFn`s to `PTransform`s as per @dhalperi style recommendations.

dhalperi:
Looks like you've subsumed thsi in #69

PR #249, opened 2016-05-10
Return null evaluators from Unavailable Reads

tgroh:
Null TransformEvaluators for sources represent a source where all splits
are currently in use or completed.

Update TransformExecutor to handle null evaluators properly.

Change TransformExecutor to a Runnable.

Backports [Beam #310](https://github.com/apache/incubator-beam/pull/310)


tgroh:
R: @kennknowles

PR #251, opened 2016-05-10
Backport incubator-beam PR #321

lukecwik:
https://github.com/apache/incubator-beam/pull/321

Note that maven archetypes already depends on guava 19.0

Also, note that I explicitly bring in google-auth-library-oauth2-http so it matches the incubator-beam version. Normally bigtable client brings in version 0.3.1


lukecwik:
R: @dhalperi

PR #253, opened 2016-05-11
Reuse DoFns in ParDoEvaluators

tgroh:
This allows the runner to avoid cloning DoFns for every input bundle.

Backports [Beam #320](https://github.com/apache/incubator-beam/pull/320)


dhalperi:
@tgroh @kennknowles Seems like there was a backporting error -- this test got shoved in the `main` subfolder not the `test` subfolder:

`DataflowJavaSDK/sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/SerializableCloningThreadLocalCacheLoaderTest.java`

dhalperi:
And `DataflowJavaSDK/sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/ThreadLocalInvalidatingTransformEvaluatorTest.java`

robertwb:
Is this change backwards compatible?

dhalperi:
@robertwb Thanks for catching this!

I think it is technically backwards compatible (no existing code breaks) but it no longer executes the Dataflow 1.x model. So @tgroh will revert.

Issue #148, opened 2016-03-15
BigTableIO is not supported in Streaming mode


ankurcha:
The class `com.google.cloud.dataflow.sdk.io.bigtable.BigtableIO.Write` is currently implemented using  `Sink`. This means that they may only be used in batch mode. It would be nice to have it implemented in a way to support streaming mode too.

An exception to that effect is:

```bash
java.lang.UnsupportedOperationException: The Write transform is not supported by the Dataflow streaming runner.
	at com.google.cloud.dataflow.sdk.runners.DataflowPipelineRunner$StreamingWrite.apply(DataflowPipelineRunner.java:2271)
	at com.google.cloud.dataflow.sdk.runners.DataflowPipelineRunner$StreamingWrite.apply(DataflowPipelineRunner.java:2262)
	at com.google.cloud.dataflow.sdk.runners.PipelineRunner.apply(PipelineRunner.java:74)
	at com.google.cloud.dataflow.sdk.runners.DataflowPipelineRunner.apply(DataflowPipelineRunner.java:413)
	at com.google.cloud.dataflow.sdk.Pipeline.applyInternal(Pipeline.java:367)
	at com.google.cloud.dataflow.sdk.Pipeline.applyTransform(Pipeline.java:274)
	at com.google.cloud.dataflow.sdk.runners.DataflowPipelineRunner.apply(DataflowPipelineRunner.java:411)
	at com.google.cloud.dataflow.sdk.Pipeline.applyInternal(Pipeline.java:367)
	at com.google.cloud.dataflow.sdk.Pipeline.applyTransform(Pipeline.java:274)
	at com.google.cloud.dataflow.sdk.values.PCollection.apply(PCollection.java:161)
	at com.google.cloud.dataflow.sdk.io.bigtable.BigtableIO$Write.apply(BigtableIO.java:401)
	at com.google.cloud.dataflow.sdk.io.bigtable.BigtableIO$Write.apply(BigtableIO.java:325)
	at com.google.cloud.dataflow.sdk.runners.PipelineRunner.apply(PipelineRunner.java:74)
	at com.google.cloud.dataflow.sdk.runners.DataflowPipelineRunner.apply(DataflowPipelineRunner.java:413)
	at com.google.cloud.dataflow.sdk.Pipeline.applyInternal(Pipeline.java:367)
	at com.google.cloud.dataflow.sdk.Pipeline.applyTransform(Pipeline.java:290)
[...]

PR #254, opened 2016-05-11
Add ShardControlledWrite override

tgroh:
This is used for TextIO and AvroIO, which provide withNumOutputShards
methods to control the number of output files. Apply this override in
the InProcessPipelineRunner.

Backports [Beam #148](https://github.com/apache/incubator-beam/pull/148)


tgroh:
R: @lukecwik

PR #255, opened 2016-05-11
Enforce non-null preconditions

tgroh:
The InProcessPipelineRunner expects some result values to be non-null.
Enforce those early.

Backports [Beam #153](https://github.com/apache/incubator-beam/pull/153)


tgroh:
R: @kennknowles

PR #257, opened 2016-05-12
BoundedReader: add getSplitPoints{Consumed,Remaining}

dhalperi:
And implement it for common sources

### DIFF
--- a/contrib/hadoop/src/main/java/com/google/cloud/dataflow/contrib/hadoop/HadoopFileSource.java
+++ b/contrib/hadoop/src/main/java/com/google/cloud/dataflow/contrib/hadoop/HadoopFileSource.java
@@ -281,6 +281,7 @@ public class HadoopFileSource<K, V> extends BoundedSource<KV<K, V>> {
     private Configuration conf;
     private RecordReader<K, V> currentReader;
     private KV<K, V> currentPair;
+    private volatile boolean done = false;
 
     /**
      * Create a {@code HadoopFileReader} based on a file or a file pattern specification.
@@ -355,6 +356,7 @@ public class HadoopFileSource<K, V> extends BoundedSource<KV<K, V>> {
           }
           // either no next split or all readers were empty
           currentPair = null;
+	  done = true;
           return false;
         }
       } catch (InterruptedException e) {
@@ -432,8 +434,8 @@ public class HadoopFileSource<K, V> extends BoundedSource<KV<K, V>> {
     }
 
     @Override
-    public final long getParallelismRemaining() {
-      if (isDone()) {
+    public final long getSplitPointsRemaining() {
+      if (done) {
         return 0;
       }
       // This source does not currently support dynamic work rebalancing, so remaining
--- a/contrib/hadoop/src/main/java/com/google/cloud/dataflow/contrib/hadoop/HadoopFileSource.java
+++ b/contrib/hadoop/src/main/java/com/google/cloud/dataflow/contrib/hadoop/HadoopFileSource.java
@@ -432,6 +432,16 @@ public class HadoopFileSource<K, V> extends BoundedSource<KV<K, V>> {
     }
 
     @Override
+    public final long getParallelismRemaining() {
+      if (isDone()) {
+        return 0;
+      }
+      // This source does not currently support dynamic work rebalancing, so remaining
+      // parallelism is always 1.
+      return 1;
+    }
+
+    @Override
     public BoundedSource<KV<K, V>> splitAtFraction(double fraction) {
       // Not yet supported. To implement this, the sizes of the splits should be used to
       // calculate the remaining splits that constitute the given fraction, then a
--- a/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BlockBasedSource.java
+++ b/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BlockBasedSource.java
@@ -206,20 +206,26 @@ public abstract class BlockBasedSource<T> extends FileBasedSource<T> {
 
     @Override
     public Double getFractionConsumed() {
-      if (getCurrentSource().getEndOffset() == Long.MAX_VALUE) {
+      if (!isStarted()) {
+        return 0.0;
+      }
+      if (isDone()) {
+        return 1.0;
+      }
+      FileBasedSource<T> source = getCurrentSource();
+      if (source.getEndOffset() == Long.MAX_VALUE) {
         return null;
       }
-      Block<T> currentBlock = getCurrentBlock();
+
       long currentBlockOffset = getCurrentBlockOffset();
-      long startOffset = getCurrentSource().getStartOffset();
-      long endOffset = getCurrentSource().getEndOffset();
+      long startOffset = source.getStartOffset();
+      long endOffset = source.getEndOffset();
       double fractionAtBlockStart =
           ((double) (currentBlockOffset - startOffset)) / (endOffset - startOffset);
       double fractionAtBlockEnd =
           ((double) (currentBlockOffset + getCurrentBlockSize() - startOffset)
               / (endOffset - startOffset));
-      double blockFraction = (currentBlock == null)
-          ? 0.0 : currentBlock.getFractionOfBlockConsumed();
+      double blockFraction = getCurrentBlock().getFractionOfBlockConsumed();
       return Math.min(
           1.0,
           fractionAtBlockStart + blockFraction * (fractionAtBlockEnd - fractionAtBlockStart));
--- a/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BlockBasedSource.java
+++ b/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BlockBasedSource.java
@@ -205,6 +205,7 @@ public abstract class BlockBasedSource<T> extends FileBasedSource<T> {
     }
 
     @Override
+    @Nullable
     public Double getFractionConsumed() {
       if (!isStarted()) {
         return 0.0;
@@ -214,6 +215,7 @@ public abstract class BlockBasedSource<T> extends FileBasedSource<T> {
       }
       FileBasedSource<T> source = getCurrentSource();
       if (source.getEndOffset() == Long.MAX_VALUE) {
+        // Unknown end offset, so we cannot tell.
         return null;
       }
 
--- a/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BlockBasedSource.java
+++ b/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BlockBasedSource.java
@@ -210,10 +210,6 @@ public abstract class BlockBasedSource<T> extends FileBasedSource<T> {
         return null;
       }
       Block<T> currentBlock = getCurrentBlock();
-      if (currentBlock == null) {
-        // There is no current block (i.e., the read has not yet begun).
-        return 0.0;
-      }
       long currentBlockOffset = getCurrentBlockOffset();
       long startOffset = getCurrentSource().getStartOffset();
       long endOffset = getCurrentSource().getEndOffset();
@@ -222,11 +218,11 @@ public abstract class BlockBasedSource<T> extends FileBasedSource<T> {
       double fractionAtBlockEnd =
           ((double) (currentBlockOffset + getCurrentBlockSize() - startOffset)
               / (endOffset - startOffset));
+      double blockFraction = (currentBlock == null)
+          ? 0.0 : currentBlock.getFractionOfBlockConsumed();
       return Math.min(
           1.0,
-          fractionAtBlockStart
-          + currentBlock.getFractionOfBlockConsumed()
-            * (fractionAtBlockEnd - fractionAtBlockStart));
+          fractionAtBlockStart + blockFraction * (fractionAtBlockEnd - fractionAtBlockStart));
     }
 
     @Override
--- a/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
+++ b/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
@@ -40,9 +40,16 @@ import javax.annotation.Nullable;
  * <li>Size estimation: {@link #getEstimatedSizeBytes};
  * <li>Telling whether or not this source produces key/value pairs in sorted order:
  * {@link #producesSortedKeys};
- * <li>The reader ({@link BoundedReader}) supports progress estimation
- * ({@link BoundedReader#getFractionConsumed}) and dynamic splitting
- * ({@link BoundedReader#splitAtFraction}).
+ * <li>The accompanying {@link BoundedReader reader} has additional functionality to enable runners
+ * to dynamically adapt based on runtime conditions.
+ *     <ul>
+ *       <li>Progress estimation ({@link BoundedReader#getFractionConsumed})
+ *       <li>Tracking of parallelism, to determine with the current source can be split
+ *        ({@link BoundedReader#getParallelismConsumed()} and
+ *        {@link BoundedReader#getParallelismRemaining()}).
+ *       <li>Dynamic splitting of the current source ({@link BoundedReader#splitAtFraction}).
+ *     </ul>
+ *     </li>
  * </ul>
  *
  * <p>To use this class for supporting your custom input type, derive your class
@@ -147,33 +154,39 @@ public abstract class BoundedSource<T> extends Source<T> {
      * Returns the total amount of parallelism in the consumed (returned and processed) range of
      * this reader's current {@link BoundedSource} (as would be returned by
      * {@link #getCurrentSource}). This corresponds to all split point records (see
-     * {@link RangeTracker}) returned by this reader, <em>excluding</em> the current record.
+     * {@link RangeTracker}) returned by this reader, <em>excluding</em> the last split point
+     * returned if the reader is not finished.
      *
-     * <p>Consider the following two examples. An input that can be read in parallel down to the
+     * <p>Consider the following examples: (1) An input that can be read in parallel down to the
      * individual records, such as {@link CountingSource#upTo}, is called "perfectly splittable".
-     * Conversely, an example of a non-perfectly splittable input is a block-compressed file format
-     * such as {@link AvroIO}, in which a block of records has to be read as a whole, but different
-     * blocks can be read in parallel.
+     * (2) a "block-compressed" file format such as {@link AvroIO}, in which a block of records has
+     * to be read as a whole, but different blocks can be read in parallel. (3) An "unsplittable"
+     * input such as a cursor in a database.
      *
      * <ul>
      * <li>Any {@link BoundedReader reader} that is unstarted (aka, has never had a call to
-     * {@link #start}) has a consumed parallelism of 0.
+     * {@link #start}) has a consumed parallelism of 0. This condition holds independent of whether
+     * the input is splittable.
      * <li>Any {@link BoundedReader reader} that has only returned its first element (aka,
      * has never had a call to {@link #advance}) has a consumed parallelism of 0: the first element
-     * is the current element and is still being processed.
-     * <li>When processing record #30 (starting at 1) out of 50 in a perfectly splittable 50-record
-     * input, this value should be 29.
-     * <li>In a block-compressed value consisting of 5 blocks, the value should stay at 0 until the
-     * first record of the second block is returned; stay at 1 until the first record of the third
-     * block is returned, etc. Once the end-of-file is reached then the fifth block has been
-     * consumed and the value should stay at 5.
+     * is the current element and is still being processed. This condition holds independent of
+     * whether the input is splittable.
      * <li>For an empty reader (in which the call to {@link #start} returned false), the
-     * consumed parallelism is 0.
+     * consumed parallelism is 0. This condition holds independent of whether the input is
+     * splittable.
      * <li>For a non-empty, finished reader (in which the call to {@link #start} returned true and
      * a call to {@link #advance} has returned false), the value returned must be at least 1
-     * and should equal the total parallelism in the source. For the two examples above, the
-     * 50-record file should have a consumed parallelism of 50 and the 5-block file should have
-     * a consumed parallelism of 5.
+     * and should equal the total parallelism in the source.
+     * <li>For example (1): After returning record #30 (starting at 1) out of 50 in a perfectly
+     * splittable 50-record input, this value should be 29. When finished, the consumed parallelism
+     * should be 50.
+     * <li>For example (2): In a block-compressed value consisting of 5 blocks, the value should
+     * stay at 0 until the first record of the second block is returned; stay at 1 until the first
+     * record of the third block is returned, etc. Only once the end-of-file is reached then the
+     * fifth block has been consumed and the value should stay at 5.
+     * <li>For example (3): For any non-empty unsplittable input, the consumed parallelism is 0
+     * until the reader is finished (because the last call to {@link #advance} returned false, at
+     * which point it becomes 1.
      * </ul>
      *
      * <p>A reader that is implemented using a {@link RangeTracker} is encouraged to use the
@@ -195,34 +208,42 @@ public abstract class BoundedSource<T> extends Source<T> {
     /**
      * Returns the total amount of parallelism in the unprocessed part of this reader's current
      * {@link BoundedSource} (as would be returned by {@link #getCurrentSource}). This corresponds
-     * to all unprocessed split point records (see {@link RangeTracker}), including the current
-     * record, in the remainder part of the source.
+     * to all unprocessed split point records (see {@link RangeTracker}), including the last
+     * split point returned, in the remainder part of the source.
      *
      * <p>This function should be implemented only <strong>in addition to
      * {@link #getParallelismConsumed()}</strong> and only if <em>an exact value can be
      * returned</em>.
      *
-     * <p>Consider the following two examples. An input that can be read in parallel down to the
+     * <p>Consider the following examples: (1) An input that can be read in parallel down to the
      * individual records, such as {@link CountingSource#upTo}, is called "perfectly splittable".
-     * Conversely, an example of a non-perfectly splittable input is a block-compressed file format
-     * such as {@link AvroIO}, in which a block of records has to be read as a whole, but different
-     * blocks can be read in parallel.
+     * (2) a "block-compressed" file format such as {@link AvroIO}, in which a block of records has
+     * to be read as a whole, but different blocks can be read in parallel. (3) An "unsplittable"
+     * input such as a cursor in a database.
+     *
+     * <p>Assume for examples (1) and (2) that the number of records or blocks remaining is known:
      *
      * <ul>
      * <li>Any {@link BoundedReader reader} for which the last call to {@link #start} or
-     * {@link #advance} has returned true should return a value of at least 1 to indicate the source
-     * that it is currently reading.
-     * <li>A finished reader for which {@link #start} or {@link #advance} has returned false
-     * should return a value of 0.
-     * <li>For any source that cannot be split in general, the reader should return 1.
-     * <li>When processing record #30 (starting at 1) out of 50 in a perfectly splittable 50-record
-     * input, this value should be 21 (20 remaining + 1 current).
-     * <li>If we are reading through block 3 in a block-compressed file consisting of 5 blocks,
-     * this value should be 3 (since blocks 4 and 5 can be processed in parallel by new readers
-     * produced via dynamic work rebalancing, while the current reader remains processing block 3).
-     * <li>If we are reading through the last block in a block-compressed file, or reading or
-     * processing the last record in a perfectly splittable input, this value should be 1: apart
-     * from the current task, no additional remainder can be split off.
+     * {@link #advance} has returned true should should not return 0, because this reader itself
+     * represents parallelism at least 1. This condition holds independent of whether the input is
+     * splittable.
+     * <li>A finished reader (for which {@link #start} or {@link #advance}) has returned false
+     * should return a value of 0. This condition holds independent of whether the input is
+     * splittable.
+     * <li>For example 1: After returning record #30 (starting at 1) out of 50 in a perfectly
+     * splittable 50-record input, this value should be 21 (20 remaining + 1 current) if the total
+     * number of records is known.
+     * <li>For example 2: After returning a record in block 3 in a block-compressed file
+     * consisting of 5 blocks, this value should be 3 (since blocks 4 and 5 can be processed in
+     * parallel by new readers produced via dynamic work rebalancing, while the current reader
+     * continues processing block 3) if the total number of blocks is known.
+     * <li>For example (3): a reader for any non-empty unsplittable input, should return 1 until
+     * it is finished, at which point it should return 0.
+     * <li>For any reader: After returning the last split point in a file (e.g., the last record
+     * in example (1), the first record in the last block for example (2), or the first record in
+     * the file for example (3), this value should be 1: apart from the current task, no additional
+     * remainder can be split off.
      * </ul>
      *
      * <p>Defaults to {@link #PARALLELISM_UNKNOWN}. Any value less than 0 will be interpreted as
--- a/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
+++ b/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
@@ -28,6 +28,8 @@ import java.io.IOException;
 import java.util.List;
 import java.util.NoSuchElementException;
 
+import javax.annotation.Nullable;
+
 /**
  * A {@link Source} that reads a finite amount of input and, because of that, supports
  * some additional operations.
@@ -130,6 +132,7 @@ public abstract class BoundedSource<T> extends Source<T> {
      * methods (including itself), and it is therefore critical for it to be implemented
      * in a thread-safe way.
      */
+    @Nullable
     public Double getFractionConsumed() {
       return null;
     }
@@ -364,6 +367,7 @@ public abstract class BoundedSource<T> extends Source<T> {
      *
      * <p>By default, returns null to indicate that splitting is not possible.
      */
+    @Nullable
     public BoundedSource<T> splitAtFraction(double fraction) {
       return null;
     }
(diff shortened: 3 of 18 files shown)


sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BlockBasedSource.java
```diff
@@ -174,6 +174,15 @@ public final T getCurrent() throws NoSuchElementException {
       return currentBlock.getCurrentRecord();
     }
 
+    @Override
+    public synchronized long getParallelismRemaining() {
+      if (getCurrentBlockOffset() + getCurrentBlockSize() >= getCurrentSource().getEndOffset()) {
+        // If this block is known to be the last block in the range, no remaining parallelism.
```
fahndrich: Remaining parallelism is 1, not "no remaining parallelism".

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
```diff
@@ -133,6 +133,95 @@ public Double getFractionConsumed() {
     }
 
     /**
+     * A constant to use as the return value for {@link #getParallelismConsumed()} or
+     * {@link #getParallelismRemaining()} when the exact value is unknown.
+     */
+    public static final long PARALLELISM_UNKNOWN = -1;
```
fahndrich: Why are we going with long instead of Double, where we can represent NaN and Inf?

dhalperi: Consumed parallelism: clearly integral.

Remaining paralellism: use is exhorted to only populate if exact. Hence also clearly integral.

For the user API, these concepts are integral and neither of these can be `Inf`.

Of course the interface between the job and the Dataflow service need not follow the user API.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
```diff
@@ -133,6 +133,95 @@ public Double getFractionConsumed() {
...
+     * <li>For a non-empty, finished reader (in which the call to {@link #start} returned true and
+     * a call to {@link #advance} has returned false), the value returned is at must be at least 1
+     * and should equal the total parallelism in the source. For the two examples above, the
+     * 50-record file should have a consumed parallelism of 50.
+     * </ul>
+     *
+     * <p>A reader that is implemented using a {@link org.apache.beam.sdk.io.range.RangeTracker} is
+     * encouraged to delegate this function to its {@code getParallelismConsumed} method, if
+     * implemented.
+     *
```
fahndrich: "its" -> the RangeTracker's.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
```diff
@@ -133,6 +133,95 @@ public Double getFractionConsumed() {
...
+     * <p>Defaults to {@link #PARALLELISM_UNKNOWN}.
+     *
+     * @see #getParallelismRemaining()
+     */
+    public long getParallelismConsumed() {
+      return PARALLELISM_UNKNOWN;
```
fahndrich: Ditto about double vs long

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
```diff
@@ -133,6 +133,95 @@ public Double getFractionConsumed() {
...
+    }
+
+    /**
+     * Returns the total amount of parallelism in the unread range of this reader's current
```
fahndrich: Unread isn't correct here, as the currently processing element is part of remaining parallelism and has been read.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
```diff
@@ -133,6 +133,95 @@ public Double getFractionConsumed() {
...
+     * <li>Any {@link BoundedReader reader} should return a value of at least 1 indicating the
+     * source it is currently reading.
+     * <li>When processing record #30 (starting at 1) out of 50 in a perfectly splittable 50-record
+     * input, this value should be 21 (20 remaining + 1 current).
+     * <li>If we are reading through block 3 in a block-compressed file consisting of 5 blocks,
+     * this value should be 3 (since blocks 4 and 5 can be processed in parallel by new readers
+     * produced via dynamic work rebalancing, while the current reader remains processing block 3).
+     * <li>If we are reading through the last block in a block-compressed file, or reading or
+     * processing the last record in a perfectly splittable input, this value should be 1: apart
+     * from the current task, no additional remainder can be split off.
```
fahndrich: Add that if the source is not splittable in general, the returned value should be 1.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
```diff
@@ -133,6 +133,95 @@ public Double getFractionConsumed() {
...
+     * </ul>
+     *
+     * <p>Defaults to {@link #PARALLELISM_UNKNOWN}.
+     *
+     * @see #getParallelismConsumed()
+     */
+    public long getParallelismRemaining() {
```
fahndrich: Ditto long vs double

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/CompressedSource.java
```diff
@@ -375,11 +384,34 @@ public T getCurrent() throws NoSuchElementException {
...
     }
 
+    @Override
+    public final long getParallelismConsumed() {
+      if (splittable) {
+        return readerDelegate.getParallelismConsumed();
+      } else if (done) {
+        return 1;
+      }
+      return 0;
```
fahndrich: This code is wrong. Last return is unreachable. If !splittable, consumed is 0, not 1.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/CompressedSource.java
```diff
@@ -375,11 +384,34 @@ public T getCurrent() throws NoSuchElementException {
...
+    }
```
fahndrich: Maybe put the last return in an else to make it easier to read.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/CompressedSource.java
```diff
@@ -375,11 +384,34 @@ public T getCurrent() throws NoSuchElementException {
...
+
+    @Override
+    public final long getParallelismRemaining() {
+      if (splittable) {
+        return readerDelegate.getParallelismRemaining();
+      } else if (done) {
+        return 0;
+      }
+      return 1;
```
fahndrich: Ditto.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/DatastoreIO.java
```diff
@@ -868,6 +868,8 @@ public DatastoreWriteResult(long recordsWritten) {
      */
     private int userLimit;
 
+    private volatile boolean done = false;
```
fahndrich: Comment in new member variables "done"?

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/OffsetBasedSource.java
```diff
@@ -264,7 +264,13 @@ public final boolean start() throws IOException {
 
     @Override
     public final boolean advance() throws IOException {
-      return advanceImpl() && rangeTracker.tryReturnRecordAt(isAtSplitPoint(), getCurrentOffset());
```
fahndrich: Is there somewhere to document the change in "pattern"? I remember seeing the above pattern as the "suggested" way to do readers.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/OffsetBasedSource.java
```diff
@@ -308,6 +314,19 @@ public Double getFractionConsumed() {
     }
 
     @Override
+    public long getParallelismConsumed() {
+      return rangeTracker.getProcessedSplitPoints();
+    }
+
+    @Override
+    public long getParallelismRemaining() {
+      if (getCurrentOffset() + 1 >= rangeTracker.getStopPosition()) {
```
fahndrich: Why + 1? Typically, if getCurrentOffset == getStopPosition, we would be done (0), so the fact that we are not done would indicate already 1. This source could also track done to return 0.

fahndrich: Given that you have (getCurrentSource().allowsDynamicSplitting(), shouldn't you report 1 here in case the source does not?

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/range/OffsetRangeTracker.java
```diff
@@ -76,8 +88,14 @@ public synchronized boolean tryReturnRecordAt(boolean isAtSplitPoint, long recor
               recordStart,
               lastRecordStart));
     }
+
+    if (isAtSplitPoint && lastRecordStart >= -1) {
+      ++seenSplitPoints;
```
fahndrich: lastRecordStart is always >= -1

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
fahndrich: The method seems to be called getProcessedSplitPoints. Make it consistent either way.

sdk/src/test/java/com/google/cloud/dataflow/sdk/io/CompressedSourceTest.java
```diff
@@ -466,6 +466,7 @@ public boolean producesSortedKeys(PipelineOptions options) throws Exception {
 
       public ByteReader(ByteSource source) {
         super(source);
+        offset = source.getStartOffset() - 1;
```
fahndrich: ???

fahndrich:
Your change is missing the changes to RangeTracker itself.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
dhalperi: Updated the specific comment to be more accurate. Specifically, suggests using a range tracker's ability to count split points and then link to `OffsetBasedReader`. We can't actually force a change to the RangeTracker API (backwards incompatible), so instead it makes sense to give a concrete example.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/OffsetBasedSource.java
dhalperi: The current position can be >= the stop position when we're emitting non-split-point records. I assume it's useful to know that we're on the last split.

As for why +1: this is to determine if we're at the last record. If we're on record 99 and the stop offset is 100, we did not meet the conditions you described. The only way I know of to distinguish 99 from 98 (or 1) is to + 1 for the check.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/range/OffsetRangeTracker.java
dhalperi: Yes, there are no new tests yet ;) -- this was intended for high-level feedback.

contrib/hadoop/src/main/java/com/google/cloud/dataflow/contrib/hadoop/HadoopFileSource.java
```diff
@@ -432,6 +432,12 @@ private Double getProgress() {
     }
 
     @Override
+    public final long getParallelismRemaining() {
+      // Not yet supported. See #splitAtFraction.
```
jkff: Better to say "This source currently doesn't support dynamic work rebalancing, so remaining parallelism is always 1".

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
jkff: Can we return a Long instead of a long then? null seems a good way to represent unknown.

sdk/src/test/java/com/google/cloud/dataflow/sdk/io/CompressedSourceTest.java
dhalperi: We were always defaulting to -1, which is wrong if the range does not start at 0.

As for why -1 -- the easiest implementation here is to increment offset in every call to read a record, so they all start 1 before the first offset so that after the first record, getCurrentOffset is accurate.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
```diff
@@ -133,6 +133,95 @@ public Double getFractionConsumed() {
...
+     * {@link #getParallelismRemaining()} when the exact value is unknown.
+     */
+    public static final long PARALLELISM_UNKNOWN = -1;
+
+    /**
+     * Returns the total amount of parallelism in the consumed (and processed) range of this
+     * reader's current {@link BoundedSource} (as would be returned by {@link #getCurrentSource}).
+     *
+     * <p>The "parallelism" consumed does not necessarily equal the number of <em>records</em>
+     * output and processed, but rather the number of distinct (ideally non-empty) <em>sources</em>
```
jkff: This description seems vague, but consumed parallelism is well-defined and unambiguous - it's the number of split points returned so far - 1 (or alternatively, number of processed "blocks" - ranges of records between two split points). I think it's best to refer here to RangeTracker for definition of split points.

dhalperi:
There are no changes to RangeTracker itself, except maybe javadoc? We can't actually add a new required function, because it's an interface.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
```diff
@@ -133,6 +133,95 @@ public Double getFractionConsumed() {
...
 
     /**
+     * A constant to use as the return value for {@link #getParallelismConsumed()} or
+     * {@link #getParallelismRemaining()} when the exact value is unknown.
+     */
+    public static final long PARALLELISM_UNKNOWN = -1;
+
+    /**
+     * Returns the total amount of parallelism in the consumed (and processed) range of this
+     * reader's current {@link BoundedSource} (as would be returned by {@link #getCurrentSource}).
```
jkff: I'd suggest to explain the difference between consumed and processed here, or refer only to processed. An unprepared reader might assume that these are interchangeable, but they aren't.

I suggest the following definitions:
- A record is "returned" (consumed?) via a call to start/advance that returns true.
- A record is "processed" when the next call to start/advance is made, because we assume that processing happens via a read/process/read/process/.../EOF loop.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
```diff
@@ -133,6 +133,96 @@ public Double getFractionConsumed() {
...
+     * <p>Consider the following two examples. An input that can be read in parallel down to the
+     * individual records, such as {@link CountingSource#upTo}, is called "perfectly splittable".
+     * Conversely, an example of a non-perfectly splittable input is a block-compressed file format
+     * such as {@link AvroIO}, in which a block of records has to be read as a whole, but different
+     * blocks can be read in parallel.
+     *
+     * <ul>
+     * <li>Any {@link BoundedReader reader} that is unstarted (aka, has never had a call to
+     * {@link #start}) has a consumed parallelism of 0.
+     * <li>Any {@link BoundedReader reader} that is currently processing its first element (aka,
```
jkff: This is a bit confusing because the reader is not _processing_ the element. Better just say "that has not yet returned its second element" and remove "and thus has not been consumed".

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
```diff
@@ -133,6 +133,96 @@ public Double getFractionConsumed() {
...
+     * has never had a call to {@link #advance}) has a consumed parallelism of 0: the first element
+     * is still being processed, and thus has not been consumed.
+     * <li>When processing record #30 (starting at 1) out of 50 in a perfectly splittable 50-record
+     * input, this value should be 29.
+     * <li>If we are reading through block 3 in a block-compressed file consisting of 5 blocks,
```
jkff: This merits more clarification. I suggest this rephrasing.
In a block-compressed file consisting of 5 blocks, the value should stay at 0 until the first record of the second block is returned; stay at 1 until the first record of the third block is returned; etc.; stay at 4 from when the first record of the 5th block is returned until EOF; then stay at 5.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
```diff
@@ -133,6 +133,96 @@ public Double getFractionConsumed() {
...
+     * this value should be 2.
+     * <li>For an empty reader (in which the call to {@link #start} returned false), the
+     * consumed parallelism is 0.
+     * <li>For a non-empty, finished reader (in which the call to {@link #start} returned true and
+     * a call to {@link #advance} has returned false), the value returned is at must be at least 1
```
jkff: remove "is at"

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
```diff
@@ -133,6 +133,96 @@ public Double getFractionConsumed() {
...
+      return PARALLELISM_UNKNOWN;
+    }
+
+    /**
+     * Returns the total amount of parallelism in the unprocessed part of this reader's current
+     * {@link BoundedSource} (as would be returned by {@link #getCurrentSource}). This function
+     * should be implemented only <strong>in addition to {@link #getParallelismConsumed()}</strong>
+     * and only if <em>an exact value can be returned</em>.
+     *
+     * <p>The "parallelism" remaining does not necessarily equal the number of <em>records</em>,
```
jkff: Same comments about using a precise definition.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
jkff: Unless it's empty / at EOF. But in that case it's "okay" to return 1 anyway and assume that upstream code will correct this.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/CompressedSource.java
```diff
@@ -356,14 +356,23 @@ public final DecompressingChannelFactory getChannelFactory() {
 
     private final FileBasedReader<T> readerDelegate;
     private final CompressedSource<T> source;
+    private final boolean splittable;
```
jkff: I'd prefer names isSplittable, isDone

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
```diff
@@ -133,6 +133,96 @@ public Double getFractionConsumed() {
...
+     * </ul>
+     *
+     * <p>A reader that is implemented using a {@code RangeTracker} is encouraged to use the
+     * range tracker's ability to count split points to implement this method. See
+     * {@link OffsetBasedSource.OffsetBasedReader} for an example.
+     *
+     * <p>Defaults to {@link #PARALLELISM_UNKNOWN}.
+     *
+     * @see #getParallelismRemaining()
+     */
```
jkff: Comment on thread safety of this and the other function? (I suppose that's why you have those volatile variables below?)

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/DatastoreIO.java
```diff
@@ -888,6 +890,21 @@ public Entity getCurrent() {
     }
 
     @Override
+    public final long getParallelismConsumed() {
+      if (done) {
```
jkff: use ?:

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/OffsetBasedSource.java
```diff
@@ -264,7 +264,18 @@ public final boolean start() throws IOException {
 
     @Override
     public final boolean advance() throws IOException {
-      return advanceImpl() && rangeTracker.tryReturnRecordAt(isAtSplitPoint(), getCurrentOffset());
+      if (!advanceImpl()) {
+        // Reader could not read another record. Add an extra split point in order to "finish" the
```
jkff: I'd prefer not to reuse tryReturnRecordAt for this purpose - better add a method that specifically tells the range tracker that EOF was reached. This is not an extra split point because split points are existing records at split point positions, and there is no record here. We're just telling the range tracker that the *previous* split point (and its block) has been processed so it's okay to increment the counter.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/OffsetBasedSource.java
```diff
@@ -308,6 +319,25 @@ public Double getFractionConsumed() {
     }
 
     @Override
+    public long getParallelismConsumed() {
+      return rangeTracker.getProcessedSplitPoints();
```
jkff: Would be good to be consistent with getParallelismConsumed: getSplitPointsProcessed?

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/OffsetBasedSource.java
```diff
@@ -308,6 +319,25 @@ public Double getFractionConsumed() {
...
+    }
+
+    @Override
+    public long getParallelismRemaining() {
+      if (done) {
```
jkff: I suggest to get rid of "done" here too, as well as everywhere else.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/range/OffsetRangeTracker.java
```diff
@@ -68,6 +71,15 @@ public synchronized boolean tryReturnRecordAt(boolean isAtSplitPoint, long recor
       throw new IllegalStateException(
           String.format("The first record [starting at %d] must be at a split point", recordStart));
     }
+    if (recordStart < startOffset) {
+      throw new IllegalStateException(
+          String.format(
+              "Trying to return record [starting at %d] "
+                  + "which is before the start offset [starting at %d]",
```
jkff: [starting at] is unnecessary when logging about startOffset.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/range/OffsetRangeTracker.java
```diff
@@ -155,6 +172,25 @@ public synchronized double getFractionConsumed() {
...
+   * offset.
+   *
+   * <p>Note that for correctness when implementing {@link BoundedReader#getParallelismConsumed()},
+   * if a reader finishes before {@link #tryReturnRecordAt(boolean, long)} returns false,
+   * the reader should add an additional call to {@link #tryReturnRecordAt(boolean, long)} with a
+   * large position (such as {@link Long#MAX_VALUE}). This will indicate that processing for the
+   * last seen split point has been finished.
+   */
+  public synchronized long getProcessedSplitPoints() {
+    return seenSplitPoints;
```
jkff: Rename to splitPointsProcessed?

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/range/OffsetRangeTracker.java
```diff
@@ -76,8 +88,14 @@ public synchronized boolean tryReturnRecordAt(boolean isAtSplitPoint, long recor
               recordStart,
               lastRecordStart));
     }
+
+    if (isAtSplitPoint && lastRecordStart > -1) {
+      ++seenSplitPoints;
```
jkff: Add a comment here, saying that we're incrementing splitPointsProcessed on behalf of the *previous* split point, because we've seen the next one (which means the previous one has been processed)

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/range/OffsetRangeTracker.java
```diff
@@ -155,6 +172,25 @@ public synchronized double getFractionConsumed() {
...
+   *
+   * <p>A split point at a particular offset has been seen if there has been a corresponding call
+   * to {@link #tryReturnRecordAt(boolean, long)} with {@code isAtSplitPoint} true. It has been
+   * processed if there has been a <em>subsequent</em> call to
+   * {@link #tryReturnRecordAt(boolean, long)} with {@code isAtSplitPoint} true and at a larger
+   * offset.
+   *
+   * <p>Note that for correctness when implementing {@link BoundedReader#getParallelismConsumed()},
+   * if a reader finishes before {@link #tryReturnRecordAt(boolean, long)} returns false,
+   * the reader should add an additional call to {@link #tryReturnRecordAt(boolean, long)} with a
```
jkff: Change this paragraph after adding that extra reportEOF method

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
dhalperi: We can... but since the caller already has to check that the value is non-negative (and interpret that as UNKNOWN) it seemed simpler to include that behavior naturally. Neither way is very pretty, but java doesn't have unsigned values.

contrib/hadoop/src/main/java/com/google/cloud/dataflow/contrib/hadoop/HadoopFileSource.java
dhalperi: done.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
dhalperi: done

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
dhalperi: clarified

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
dhalperi: clarified

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
dhalperi: done

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
dhalperi: done, thanks.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
dhalperi: removed this para

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/CompressedSource.java
dhalperi: Java convention is that `isSplittable` is the getter for a boolean named splittable.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/DatastoreIO.java
dhalperi: I am under the impression that it's important that consumed return 1 when finished, to distinguish it from empty. So given that we have to track doneness, it's not really extra work to return 0 when done in remaining.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/OffsetBasedSource.java
dhalperi: I thought about that approach, but I decided that this one makes most sense. For example, we'll get an error if the user calls this twice with the same value (vs adding a boolean to track EOF called twice).

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/OffsetBasedSource.java
dhalperi: done. thanks.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/range/OffsetRangeTracker.java
dhalperi: done

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/range/OffsetRangeTracker.java
dhalperi: done

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/range/OffsetRangeTracker.java
dhalperi: done

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
dhalperi: done

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/CompressedSource.java
```diff
@@ -375,11 +384,32 @@ public T getCurrent() throws NoSuchElementException {
       return readerDelegate.getCurrent();
     }
 
+    @Override
+    public final long getParallelismConsumed() {
+      if (splittable) {
+        return readerDelegate.getParallelismConsumed();
+      }
+
+      return done ? 1 : 0;
```
dhalperi: if (done && numRecordsRead > 0) { 1} else 0

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
```diff
@@ -133,6 +135,107 @@ public Double getFractionConsumed() {
...
+    public long getParallelismConsumed() {
+      return PARALLELISM_UNKNOWN;
+    }
+
+    /**
+     * Returns the total amount of parallelism in the unprocessed part of this reader's current
+     * {@link BoundedSource} (as would be returned by {@link #getCurrentSource}). This corresponds
+     * to all unprocessed split point records (see {@link RangeTracker}), including the current
+     * record, in the remainder part of the source.
+     *
```
jkff: Unnecessary blank line, and unnecessary </p> in the next paragraph (<p> is self-closing).

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
```diff
@@ -133,6 +135,107 @@ public Double getFractionConsumed() {
...
+     * individual records, such as {@link CountingSource#upTo}, is called "perfectly splittable".
+     * Conversely, an example of a non-perfectly splittable input is a block-compressed file format
+     * such as {@link AvroIO}, in which a block of records has to be read as a whole, but different
+     * blocks can be read in parallel.
+     *
+     * <ul>
+     * <li>Any {@link BoundedReader reader} for which the last call to {@link #start} or
+     * {@link #advance} has returned true should return a value of at least 1 to indicate the source
+     * that it is currently reading.
+     * <li>A finished reader for which {@link #start} or {@link #advance} has returned false and
```
jkff: remove "and"

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/CompressedSource.java
```diff
@@ -424,7 +455,10 @@ protected final boolean readNextRecord() throws IOException {
      */
     @Override
     protected final long getCurrentOffset() {
-      return readerDelegate.getCurrentOffset();
+      if (splittable) {
```
jkff: Why this change? getCurrentOffset() is documented to return offset of the current record, rather than the last split point before it.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/CompressedSource.java
jkff: Is this still TBD?

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/OffsetBasedSource.java
jkff: Maybe create a method reportEOF() that would be implemented as tryReturnRecordAt() ? I'd prefer if at least the calling code did not conflate the concepts of EOF and returning a record.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/CompressedSource.java
dhalperi: Here's my concern: the reader's currentOffset is unrelated to the offsets in the dimension that the source, its size, and the way that the service reads them. The reader is getting a decompressed stream, so the source could be 100B big and the currentOffset could go to 100M.

Maybe this is not a bug -- progress will just quickly go to 100% and stay there? -- but seemed undesirable.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
dhalperi: done

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BoundedSource.java
dhalperi: done thanks

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/CompressedSource.java
dhalperi: yes thanks. pending a test.

PR #266, opened 2016-05-15
SerializableCloningThreadLocalCacheLoaderTest: move to correct directory

dhalperi:
Should be in test, not main.


dhalperi:
Merging to unblock myself. future R: @tgroh @kennknowles

PR #268, opened 2016-05-16
BatchBigQueryIONativeReadTranslator: fixup format string

dhalperi:
Also, inline the constant and deprecate the one in BigQueryIO.java


dhalperi:
R: @peihe

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/DataflowPipelineRunner.java
```diff
@@ -2175,8 +2175,13 @@ private void translateWriteHelper(
           // If user does not specify a project we assume the table to be located in the project
           // that owns the Dataflow job.
           String projectIdFromOptions = context.getPipelineOptions().getProject();
-          LOG.warn(String.format(BigQueryIO.SET_PROJECT_FROM_OPTIONS_WARNING, table.getDatasetId(),
-              table.getDatasetId(), table.getTableId(), projectIdFromOptions));
```
dhalperi: in case it's not obvious after reformatting, the bug was in the double use of `table.getDatasetId()`.

davorbonaci:
LGTM

PR #276, opened 2016-05-19
Delete a copy of GroupByKey introduced erroneously

dhalperi:
And blacklist `org.apache.beam` from import to catch in future.


swegner:
LGTM!

dhalperi:
As Scott pointed out, this checkstyle will not catch files not on the build path. But it would catch bad imports still left in (although those should be caught by compiling)

PR #284, opened 2016-05-23
Change-detector test for PubsubIO translation

kennknowles:
These tests were written against the v1.5.0 tag, so they ensure that subsequent releases do not deviate from the translation there.


kennknowles:
There were conflicts in the import statements. In order to keep my branch rooted at 1.5.0, I added a pre-emptive merge resolution (which violates checkstyle but is isolated to a single commit) rather than rebasing.

kennknowles:
There may be other relevant fields to set up in the tests; I just did those that seemed obvious and sufficient. LMK.

PR #285, opened 2016-05-23
Limited parallelism fixup

dhalperi:
Two minor changes caught by internal builds


dhalperi:
https://github.com/apache/incubator-beam/pull/373

PR #293, opened 2016-06-03
Reduce Visibility of InProcessPipelineRunner implementations

tgroh:
Only publicly used elements are visible.

Remove javadoc exclusion for the runner.


tgroh:
@dhalperi @davorbonaci

PR #295, opened 2016-06-06
Update docs on per-window tables

munnerz:
We've been using per-window tables using the described method for a while now in batch pipelines with no issues. I'm assuming this doc is out of date, so proposing this change.


dhalperi:
Hi James,

Thanks for the contribution, and thanks for your patience.

Though they may work reasonably well for your use case, per-window tables don't work well enough in Dataflow batch jobs that we want to guarantee them with the same level of support as in streaming. So I'll close this PR without merging for now.

Thanks!
Dan

PR #296, opened 2016-06-06
Fix validation in PubsubIO

dhalperi:


dhalperi:
This is a backport of apache/incubator-beam#421

PR #299, opened 2016-06-08
Use a DirectExecutor for Watermark Callbacks

tgroh:
This fixes a resource leak where the executor service is not properly
shut down with the rest of the InProcessPipelineRunner.


dhalperi:
LGTM

PR #303, opened 2016-06-13
Backport: Make example AddTimestampFn range deterministic

swegner:
The timestamps added in the WindowedWordCount example are currently
based on when the bundles are executed, which makes the min/max bounds
non-deterministic. This change makes the range  based on the
construction time.

(cherry picked from commit 335202a033ced6f30f1b0e5df9da047241abc750)


swegner:
R: @bjchambers

PR #304, opened 2016-06-14
CompressedSourceTest: simplify

dhalperi:
We should use random.nextBytes(buff) instead of making the array in a loop.

The code we now point to is the same as the for loop, so the test continues to pass.

Backport of https://github.com/apache/incubator-beam/pull/456


dhalperi:
R: @tgroh

tgroh:
LGTM

PR #307, opened 2016-06-15
Backport: Fix NullPointerException in AfterWatermark display data

swegner:
Window transforms register display data for the associated trigger
function by calling its .toString() method. The AfterWatermark
trigger .toString() method was not properly handling cases where
there is no late firings registered.

(cherry picked from commit e7f2f582afd48bf5e2aacb3a36471377cb927018)

### DIFF
--- a/sdk/src/main/java/com/google/cloud/dataflow/sdk/transforms/windowing/AfterWatermark.java
+++ b/sdk/src/main/java/com/google/cloud/dataflow/sdk/transforms/windowing/AfterWatermark.java
@@ -30,6 +30,8 @@ import org.joda.time.Instant;
 import java.util.List;
 import java.util.Objects;
 
+import javax.annotation.Nullable;
+
 /**
  * <p>{@code AfterWatermark} triggers fire based on progress of the system watermark. This time is a
  * lower-bound, sometimes heuristically established, on event times that have been fully processed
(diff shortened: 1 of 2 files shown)


swegner:
R: @dhalperi 
CC: @dpmills

dhalperi:
@swegner do the changes in f4a5a71 mean that Never.ever() has not been backported?

If so, @kennknowles @bjchambers should it be?

kennknowles:
It is not important from a user-facing perspective. I would backport lazily only. For example, backporting would be another legitimate reaction to the issue Scott hit.

swegner:
As f4a5a71 demosntrates, some testing is not possible with the current `NeverTrigger`. @tgroh, how difficult would it be to backport c9a1efae25455af1e47675938e296c716ec0fa0f? Is it needed for `PAssert` anyway?

swegner:
To follow-up on the above: `NeverTrigger` cannot be easily backported without de9f10c, which removes type parameters from the `Trigger` type hierarchy. de9f10c is large and I don't believe this is sufficient motivation for backporting (/cc @kennknowles @tgroh to weigh in).

kennknowles:
The reason not to backport the extraneous parameter removal is backwards-incompatibility. It won't cause incompatibility in users' trigger expressions, but would cause a problem if they every declared a variable or parameter of type `Trigger`.

I don't actually follow why removal of the type parameter is important for backporting `NeverTrigger`.

kennknowles:
The never trigger is important for multiply triggered `PAssert` so it will have to be introduced one way or another.

dhalperi:
I'll let you decide whether or when to backport the `NeverTrigger`; this PR is good as is.

PR #310, opened 2016-06-17
Return from awaitCompletion if Already Done

tgroh:
This ensures that a call to ExecutorService#awaitCompletion returns immediately
if there are no visible updates and the executor has completed. If the executor is
in this state, no additional visible updates will be published and the call will hang.

This sequence generally will not happen, as calls via InProcessPipelineResult
return if the state is already terminal, but this ensures that parallel calls to
awaitCompletion do not hang one calling thread.

Backports https://github.com/apache/incubator-beam/pull/399


tgroh:
R: @dhalperi

PR #311, opened 2016-06-17
Shutdown the InProcessPipelineRunner after Terminating Abnormally

tgroh:
Ensure that the executor service is shutdown, and the monitor is not
rescheduled, after an exception is thrown.

Backports https://github.com/apache/incubator-beam/pull/435


tgroh:
R: @dhalperi

PR #312, opened 2016-06-17
Reuse UnboundedReaders in the InProcessRunner

tgroh:
Reuse up to a point, and then discard the reader to exercise resume from
checkpoint.

Backports https://github.com/apache/incubator-beam/pull/460


tgroh:
R: @dhalperi

PR #318, opened 2016-06-28
Backport from Beam: Increase the MESSAGES_POLLING_ATTEMPTS for slow backend starts

peihe:
This is backport from: https://github.com/apache/incubator-beam/pull/549


peihe:
R: @dhalperi

PR #319, opened 2016-06-28
Add UnboundedCountingInput#withRate

tgroh:
The rate controls the speed at which UnboundedCountingInput outputs
elements. This is an aggregate rate across all instances of the
source, and thus elements will not necessarily be output "smoothly",
or within the first period. The aggregate rate, however, will be
approximately equal to the provided rate.

Add package-private CountingSource#createUnbounded() to expose the
UnboundedCountingSource type. Make UnboundedCountingSource
package-private.

Backports https://github.com/apache/incubator-beam/pull/17


tgroh:
R: @dhalperi

tgroh:
R: @bjchambers

PR #322, opened 2016-07-01
Modified range tracker to use first response seen as start key

ianzhou1:


dhalperi:
Backport of https://github.com/apache/incubator-beam/pull/440

PR #326, opened 2016-07-02
Update to bigtable-client-core-0.3.0 and use bulk writes

ianzhou1:
Generally more stable, plus bulk writes bring 5x write throughput in
batch jobs by more efficiently using the network.


dhalperi:
Did not pass, please rebase and push again to kick off another CI build.

We should also chat briefly about the new dependencies.

ianzhou1:
R: @dhalperi

dhalperi:
@ianzhou1 I think that we found out one of the two functions you added to clone builders is not needed. Can you remove it and simplify the code?

(Same for apache/incubator-beam)

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/bigtable/BigtableIO.java
```diff
@@ -1022,6 +1030,49 @@ private static String getUserAgent() {
         info.getName(),
         info.getVersion(),
         javaVersion,
-        "0.2.3" /* TODO get Bigtable client version directly from jar. */);
+        "0.3.0" /* TODO get Bigtable client version directly from jar. */);
+  }
+
+  /**
+   * A helper function to add appropriate bulk options. See
+   * <a href="https://github.com/GoogleCloudPlatform/cloud-bigtable-client/issues/899">RetryOptions
```
dhalperi: this issue link is wrong now.

sdk/src/test/java/com/google/cloud/dataflow/sdk/io/bigtable/BigtableIOTest.java
```diff
@@ -517,6 +524,58 @@ public void testGetSplitPointsConsumed() throws Exception {
     reader.close();
   }
 
+  @Test
+  public void testAddBulkOptions() {
+    BigtableOptions.Builder optionsBuilder = BIGTABLE_OPTIONS.toBuilder();
```
dhalperi: presumably set some non-defaults and then use your cloner - otherwise we can't tell whether or not things were actually cloned correctly.

sdk/src/test/java/com/google/cloud/dataflow/sdk/io/bigtable/BigtableIOTest.java
```diff
@@ -517,6 +524,58 @@ public void testGetSplitPointsConsumed() throws Exception {
...
+    assertEquals(BulkOptions.BIGTABLE_MAX_INFLIGHT_RPCS_PER_CHANNEL_DEFAULT
+        * optionsBuilder.getDataChannelCount(), bulkOptions.getMaxInflightRpcs());
+    assertEquals(BulkOptions.BIGTABLE_MAX_MEMORY_DEFAULT, bulkOptions.getMaxMemory());
+  }
+
+  @Test
+  public void testAddRetryOptions() {
+    final double delta = 0.0000001;
+    BigtableOptions.Builder optionsBuilder = BIGTABLE_OPTIONS.toBuilder();
+    optionsBuilder = BigtableIO.addRetryOptions(optionsBuilder);
```
dhalperi: ditto

PR #327, opened 2016-07-02
AvroSource: use a 64K buffer size for Snappy codec

dhalperi:
commons-compress defaults to a 32K buffer size for Snappy.

However, Avro uses xerial.snappy to write, which has a 64K buffer size.
When the buffer size is too small, decoding data from Snappy can cause
an EOF exception rather than finishing data.

This fixes BEAM-422.


dhalperi:
This is a backport of https://github.com/apache/incubator-beam/pull/583/files

dhalperi:
R: @lukecwik

lukecwik:
LGTM

PR #332, opened 2016-07-07
Clarify RequiresWindowAccess error message

rculbertson:
An error message is printed when a DoFn needs access to the window but does
not implement RequiresWindowAccess. This PR changes to the error message
to print the interface name instead of just "RequiresWindow", which makes it
more clear which interface the user should use.


dhalperi:
Thanks Ryan. Looks great and useful!

As most development has moved on to the incubating Apache Beam project, we are trying to submit fixes there first and then backport here. We don't want fixes here to "get lost" and need be re-discovered later.

Would you mind sending the same 1-liner to the Beam repo?: https://github.com/apache/incubator-beam

Once submitted there I can merge here. Or I can merge here and do the Beam-port myself, but we'd love for you to stay engaged.

Thanks!

dhalperi:
Hey Ryan, any preference on next steps? Are you willing to send a PR to Beam?

rculbertson:
I've made the same change in Beam project with this PR:

https://github.com/apache/incubator-beam/pull/638

dhalperi:
Thanks Ryan!

PR #334, opened 2016-07-08
Refactor StateSampler with parameterized schedule

bjchambers:
Encapsulate the logic to using SheduledExecutorService inside the
current default scheduler.

### DIFF
--- a/sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/worker/StateSampler.java
+++ b/sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/worker/StateSampler.java
@@ -19,6 +19,7 @@ package com.google.cloud.dataflow.sdk.util.common.worker;
 import com.google.api.client.util.NanoClock;
 import com.google.cloud.dataflow.sdk.util.common.Counter;
 import com.google.cloud.dataflow.sdk.util.common.CounterSet;
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.util.concurrent.ThreadFactoryBuilder;
 
 import java.io.Closeable;
@@ -48,27 +49,35 @@ public class StateSampler implements AutoCloseable {
    * An interface for scheduling state sampling.
    */
   public static interface SampleScheduler {
-    Closeable scheduleSampling(StateSampler sampler, long samplingPeriodMs);
+    Closeable scheduleSampling(StateSampler sampler);
   }
 
   /**
    * An implementation of {@link SampleScheduler} using a scheduled executor service and two
-   * callbacks per sample. The sampling uses a "stratified sampling" is used -- in every bucket of
+   * callbacks per sample. The sampling uses "stratified sampling" -- in every bucket of
    * samplingPeriodMs we choose a random point to sample. This prevents pathological behavior in
-   * case some states happen to occur at a similar period
+   * case some states happen to occur at a similar period.
    *
    * <p>The current implementation uses a fixed-rate timer with a period samplingPeriodMs as a
    * trampoline to a one-shot random timer which fires with a random delay within
    * samplingPeriodMs.
    */
-  private static final class DefaultScheduler implements SampleScheduler {
+  @VisibleForTesting
+  static final class DefaultScheduler implements SampleScheduler {
     private static final int NUM_EXECUTOR_THREADS = 16;
     private static final ScheduledExecutorService EXECUTOR_SERVICE =
         Executors.newScheduledThreadPool(NUM_EXECUTOR_THREADS,
             new ThreadFactoryBuilder().setDaemon(true).build());
 
+    private final long samplingPeriodMs;
+
+    @VisibleForTesting
+    DefaultScheduler(long samplingPeriodMs) {
+      this.samplingPeriodMs = samplingPeriodMs;
+    }
+
     @Override
-    public Closeable scheduleSampling(final StateSampler sampler, final long samplingPeriodMs) {
+    public Closeable scheduleSampling(final StateSampler sampler) {
       return new Closeable() {
         private ScheduledFuture<?> invocationFuture = null;
         private ScheduledFuture<?> invocationTriggerFuture = EXECUTOR_SERVICE.scheduleAtFixedRate(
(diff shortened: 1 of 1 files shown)


sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/worker/StateSampler.java
```diff
@@ -41,6 +44,92 @@
...
+  /**
+   * An implementation of {@link SampleScheduler} using a scheduled executor service and two
+   * callbacks per sample. The sampling uses a "stratified sampling" is used -- in every bucket of
+   * samplingPeriodMs we choose a random point to sample. This prevents pathological behavior in
+   * case some states happen to occur at a similar period
+   * <p>
+   * The current implementation uses a fixed-rate timer with a period samplingPeriodMs as a
+   * trampoline to a one-shot random timer which fires with a random delay within
+   * samplingPeriodMs.
+   * */
```
dhalperi: `*/`

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/worker/StateSampler.java
```diff
@@ -41,6 +44,92 @@
...
+  public static interface SampleScheduler {
+    Closeable scheduleSampling(StateSampler sampler, long samplingPeriodMs);
+  }
+
+  /**
+   * An implementation of {@link SampleScheduler} using a scheduled executor service and two
+   * callbacks per sample. The sampling uses a "stratified sampling" is used -- in every bucket of
+   * samplingPeriodMs we choose a random point to sample. This prevents pathological behavior in
+   * case some states happen to occur at a similar period
+   * <p>
```
dhalperi: newline, <p> on same line.

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/worker/StateSampler.java
```diff
@@ -41,6 +44,92 @@
...
+    private static final ScheduledExecutorService EXECUTOR_SERVICE =
+        Executors.newScheduledThreadPool(NUM_EXECUTOR_THREADS,
+            new ThreadFactoryBuilder().setDaemon(true).build());
+
+    @Override
+    public Closeable scheduleSampling(final StateSampler sampler, final long samplingPeriodMs) {
+      return new Closeable() {
+        private ScheduledFuture<?> invocationTriggerFuture = null;
+        private ScheduledFuture<?> invocationFuture = null;
+        {
```
dhalperi: ?

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/worker/StateSampler.java
```diff
@@ -41,6 +44,92 @@
...
+                  0,
+                  samplingPeriodMs,
+                  TimeUnit.MILLISECONDS);
+        }
+
+        @Override
+        public void close() throws IOException {
+          synchronized (this) {
+            if (invocationTriggerFuture != null) {
+              invocationTriggerFuture.cancel(false);
```
dhalperi: * set it to null in case of reclosing?
* is it important to close the invocationFuture if this throws? if so, handle exceptions accordingly?

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/worker/StateSampler.java
```diff
@@ -41,6 +44,92 @@
...
+        {
+          invocationTriggerFuture =
+              EXECUTOR_SERVICE.scheduleAtFixedRate(
+                  new Runnable() {
+                    @Override
+                    public void run() {
+                      long delay = ThreadLocalRandom.current().nextInt((int) samplingPeriodMs);
+                      synchronized (this) {
+                        if (invocationFuture != null) {
+                          invocationFuture.cancel(false);
```
dhalperi: same q re nulling

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/worker/StateSampler.java
bjchambers: Done

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/worker/StateSampler.java
bjchambers: Done

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/worker/StateSampler.java
bjchambers: I don't think this one is a problem since within the same synchronized block we immediately re-set invocationFuture after cancelling.

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/worker/StateSampler.java
bjchambers: Set to null in case of reclose.

I don't think it is important -- note that this is just moving code that previously was called directly in StateSampler#close into a self-contained Closeable.

lukecwik:
Instead of going with a wrapper type for a scheduler, why not create a fake or mock ScheduledExecutorService and also a fake/mock clock that allows you to control the execution and timing of each segment?

bjchambers:
Encapsulating the logic for scheduling sampling was intentional, as it allows us to use mechanisms other than ScheduledExecutorService for sampling. The current approach includes two callbacks needing to be invoked for every sample, which leads to a high variance. Other approaches that we could prototype: Using a sampling thread that wakes up periodically. Using a TimerTask and scheduling the timer to wake up. Using a periodic sample rather than random within each period.

All of these can be implemented to require only a single wake up, but not all of them use ScheduledExecutorService. Rather than mocking the implementation detail, this allows us to mock the actual behavior that we care about, and experiment with these multiple alternatives.

dhalperi:
LGTM

PR #344, opened 2016-07-22
sdk-all: do not repackage BulkOptions

dhalperi:
Also part of Bigtable public API surface


dhalperi:
FYI @rangadi

rangadi:
👍 . Thanks!

PR #345, opened 2016-07-22
Add KafkaIO to Contrib

tgroh:
KafkaIO is an Unbounded source for reading from Apache Kafka.

Backports KafkaIO from Apache Beam. See apache/incubator-beam
7b175dff8303b24cb92702db5b5a1a4691435bb1

### DIFF
--- a/contrib/kafka/pom.xml
+++ b/contrib/kafka/pom.xml
@@ -77,6 +77,15 @@
         </executions>
       </plugin>
 
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-javadoc-plugin</artifactId>
+        <version>2.10.3</version>
+        <configuration>
+          <quiet>true</quiet>
+        </configuration>
+      </plugin>
+
       <!-- Source plugin for generating source and test-source JARs. -->
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
--- a/contrib/kafka/pom.xml
+++ b/contrib/kafka/pom.xml
@@ -1,0 +1,120 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+  ~ Copyright (C) 2016 Google Inc.
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License"); you may not
+  ~ use this file except in compliance with the License. You may obtain a copy of
+  ~ the License at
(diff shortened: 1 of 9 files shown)


pom.xml
```diff
@@ -98,6 +98,7 @@
   <modules>
     <module>sdk</module>
     <module>examples</module>
+    <module>contrib/kafka</module>
```
dhalperi: This we do not want to do, I think, as it would probably break users if we don't push the contrib to Maven.

Instead, I think we actually just need to modify travis to ensure that the contrib builds.

contrib/kafka/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+                </configuration>
+            </plugin>
+        </plugins>
+    </build>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.google.cloud.dataflow</groupId>
+            <artifactId>google-cloud-dataflow-java-sdk-all</artifactId>
+            <version>${project.version}</version>
```
dhalperi: We are not going to version this with the Dataflow core -- make this a range like `[1.6.0, 2.0.0)`

contrib/kafka/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~-->
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>com.google.cloud.dataflow</groupId>
+        <artifactId>google-cloud-dataflow-java-sdk-parent</artifactId>
+        <version>1.7.0-SNAPSHOT</version>
```
dhalperi: suggest something like 0.1.0-SNAPSHOT

contrib/kafka/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+    </parent>
```
dhalperi: Should not depend on core SDK for parent. Should be standalone module like the rest of the contribs.

contrib/kafka/pom.xml
tgroh: Done.

contrib/kafka/pom.xml
tgroh: Done.

Copied most of the build from the Hadoop pom.

PR #346, opened 2016-07-24
Add template runner

sammcveety:
- Extend the functionality of jobFile via a runner that only writes the file
- Add support for non-local locations via IOChannelFactory
- Include unit testing

Extra eyes on why the failure test isn't working would be appreciated.

@dhalperi, could you please review?


sammcveety:
I signed it!

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/DataflowPipelineRunner.java
```diff
@@ -601,17 +603,35 @@ public DataflowPipelineJob run(Pipeline pipeline) {
...
+        LOG.info("Printed job specification to {}", options.getDataflowJobFile());
       } catch (IllegalStateException ex) {
-        LOG.warn("Cannot translate workflow spec to json for debug.");
-      } catch (FileNotFoundException ex) {
-        LOG.warn("Cannot create workflow spec output file.");
+        String error = "Cannot translate workflow spec to JSON.";
+        if (hooks != null && hooks.failOnJobFileWriteFailure()) {
+          throw new RuntimeException(error, ex);
+        } else {
+          LOG.warn(error);
```
dhalperi: is it worth adding `ex` here?

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/DataflowPipelineRunner.java
```diff
@@ -601,17 +603,35 @@ public DataflowPipelineJob run(Pipeline pipeline) {
...
+        }
+      } catch (IOException ex) {
+        String error =
+            String.format("Cannot create output file at {}", options.getDataflowJobFile());
+        if (hooks != null && hooks.failOnJobFileWriteFailure()) {
+          throw new RuntimeException(error, ex);
+        } else {
+          LOG.warn(error);
```
dhalperi: and here?

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/TemplatingDataflowPipelineRunner.java
```diff
@@ -0,0 +1,101 @@
...
+import com.google.common.collect.ImmutableList;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * A {@link PipelineRunner} that's like {@link DataflowPipelineRunner} but only stores a template of
+ * a job.
+ *
+ * <p>
+ * Requires that {@link getDataflowJobFile} is set.
```
dhalperi: merge into previous line?

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/TemplatingDataflowPipelineRunner.java
```diff
@@ -0,0 +1,101 @@
...
+    }
+  }
+
+  /** Constructs a runner from the provided options. */
+  public static TemplatingDataflowPipelineRunner fromOptions(PipelineOptions options) {
+    DataflowPipelineDebugOptions dataflowOptions =
+        PipelineOptionsValidator.validate(DataflowPipelineDebugOptions.class, options);
+    DataflowPipelineRunner dataflowPipelineRunner =
+        DataflowPipelineRunner.fromOptions(dataflowOptions);
+    Preconditions.checkNotNull(dataflowOptions.getDataflowJobFile());
```
dhalperi: static import the method and use it directly; add an error message if `null`.

Also, maybe us `Strings.isNullOrEmpty` to also catch the empty string instead of just `null`?

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/TemplatingDataflowPipelineRunner.java
```diff
@@ -0,0 +1,101 @@
...
+
+    return new TemplatingDataflowPipelineRunner(dataflowPipelineRunner, options);
+  }
+
+  /** {@inheritDoc} */
```
dhalperi: drop -- this is the default.

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/TemplatingDataflowPipelineRunner.java
```diff
@@ -0,0 +1,101 @@
...
+  @Override
+  public DataflowPipelineJob run(Pipeline p) {
+    dataflowPipelineRunner.setHooks(new TemplateHooks());
+    final DataflowPipelineJob job = dataflowPipelineRunner.run(p);
+    LOG.info("Template successfully created.");
+    return null;
```
dhalperi: not sure what this should be. This function is not really allowed to return null, and in fact the standard user pattern may include calling `.waitToFinish()` on it, e.g.,

@francesperry  ?

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/TemplatingDataflowPipelineRunner.java
```diff
@@ -0,0 +1,101 @@
...
+  }
+
+  @Override
+  public <OutputT extends POutput, InputT extends PInput> OutputT apply(
+      PTransform<InputT, OutputT> transform, InputT input) {
+    return dataflowPipelineRunner.apply(transform, input);
```
dhalperi: maybe add a comment that we delegate graph building to the inner runner.

sdk/src/test/java/com/google/cloud/dataflow/sdk/runners/TemplatingDataflowPipelineRunnerTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+  @Rule
+  public ExpectedException expectedThrown = ExpectedException.none();
+
+  @Rule
+  public final TemporaryFolder tmpDir = new TemporaryFolder();
+
+  /**
+   * Creates a mocked {@link DataflowPipelineJob} with the given {@code projectId} and {@code jobId}
+   * .
```
dhalperi: merge into prev, or wrap differently??

sdk/src/test/java/com/google/cloud/dataflow/sdk/runners/TemplatingDataflowPipelineRunnerTest.java
```diff
@@ -0,0 +1,122 @@
...
+  /**
+   * Tests that the {@link TemplatingDataflowPipelineRunner} throws the appropriate exception when
+   * an output file is not writable.
+   */
+  @Test
+  public void testLoggedErrorForFile() throws Exception {
+    // TODO: Determine why this isn't failing.
+    // expectedThrown.expect(IOException.class);
+    createMockRunner(createMockJob("testJobDone-projectId", "testJobDone-jobId"), "/bad/path").run(
+        DirectPipeline.createForTest());
```
dhalperi: How can we help here?

sdk/src/test/java/com/google/cloud/dataflow/sdk/runners/TemplatingDataflowPipelineRunnerTest.java
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  @Test
+  public void testToString() {
+    DataflowPipelineOptions options = PipelineOptionsFactory.as(DataflowPipelineOptions.class);
+    options.setJobName("TestJobName");
+    options.setDataflowJobFile("foo");
+    options.setProject("test-project");
+    options.setTempLocation("gs://test/temp/location");
+    options.setTempLocation("gs://test/temp/location");
```
dhalperi: drop?

dhalperi:
Minor comments. How can we help round out / debug the tests?

I'd also like @bjchambers to take a look.

sdk/src/test/java/com/google/cloud/dataflow/sdk/runners/TemplatingDataflowPipelineRunnerTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+  /**
+   * Tests that the {@link TemplatingDataflowPipelineRunner} throws the appropriate exception when
+   * an output file is not writable.
+   */
+  @Test
+  public void testLoggedErrorForFile() throws Exception {
+    // TODO: Determine why this isn't failing.
+    // expectedThrown.expect(IOException.class);
+    createMockRunner(createMockJob("testJobDone-projectId", "testJobDone-jobId"), "/bad/path").run(
```
bjchambers: I'm not sure this is the problem, but I've seen weird behaviors when using this kind of pattern to create mocks. I'd suggest:

1. Using fields instead of methods to create mocks, by annotating them with `@Mock` and using `MockitoAnnotations.initMocks` in the `@Before` `setUp` method.
2. Setting the expectations as described.

This moves most of the initialization/setup code into the `@Before`.

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/DataflowPipelineRunner.java
```diff
@@ -601,17 +603,35 @@ public DataflowPipelineJob run(Pipeline pipeline) {
...
+        String error = "Cannot translate workflow spec to JSON.";
+        if (hooks != null && hooks.failOnJobFileWriteFailure()) {
+          throw new RuntimeException(error, ex);
+        } else {
+          LOG.warn(error);
+        }
+      } catch (IOException ex) {
+        String error =
+            String.format("Cannot create output file at {}", options.getDataflowJobFile());
+        if (hooks != null && hooks.failOnJobFileWriteFailure()) {
```
bjchambers: It may be worth putting this logic (handling hooks, throwing or not, etc.) into a helper method?

sammcveety:
Ready for another round, thank you.  Mystery solved: culprit was overaggressive mocking.

sdk/src/test/java/com/google/cloud/dataflow/sdk/runners/TemplatingDataflowPipelineRunnerTest.java
```diff
@@ -0,0 +1,123 @@
...
+    expectedLogs.verifyInfo("Template successfully created");
+  }
+
+  /**
+   * Tests that the {@link TemplatingDataflowPipelineRunner} throws the appropriate exception when
+   * an output file is not writable.
+   */
+  @Test
+  public void testLoggedErrorForFile() throws Exception {
+    expectedThrown.expect(RuntimeException.class);
```
bjchambers: suggest moving the `expectedThrown.expect` down to right before the call that should throw. Especially for errors like RuntimeException, it is also helpful to add an expected message. Otherwise, if `options.setJobName(...)` decided to throw a RuntimeException this test would still pass.

sdk/src/test/java/com/google/cloud/dataflow/sdk/runners/TemplatingDataflowPipelineRunnerTest.java
```diff
@@ -0,0 +1,123 @@
...
+    MockitoAnnotations.initMocks(this);
+    when(mockJob.getProjectId()).thenReturn("project");
+    when(mockJob.getJobId()).thenReturn("job");
+  }
+
+  /**
+   * Returns a {@link TemplatingDataflowPipelineRunner} that will return the provided a job to
+   * return. Some {@link PipelineOptions} will be extracted from the job, such as the project ID.
+   */
+  private TemplatingDataflowPipelineRunner createMockRunner(
```
bjchambers: I think you could use the MockitoAnnotations to create the mock runner as well, and do more initialization in setup.

sammcveety:
Done, please take another look.

bjchambers:
Looks good from my side.

dhalperi:
I'm still concerned about returning `null` from the `run` method.

Have you tried any of the standard examples with this? What happens, e.g., if you run `WindowedWordCount`?

dhalperi:
(Does templating with a streaming job make sense?)

sammcveety:
WindowedWordCount, as written, wouldn't make sense because you wouldn't want to start the mock source right away.

Generally, yes, I think streaming templates make sense, though.

sammcveety:
How would you like to move forward on this?  I could see keeping the null behavior, or returning a vacuous DataflowPipelineJob object.  Neither seems great.

dhalperi:
I think that implementing an unhelpful `DataflowPipelineJob` object is better than making the return value of `run` `@Nullable`. We'll just have to produce some guidance about how to write your main program when we roll this feature out to ensure that users get sensible behavior.

Maybe just throw `UnsupportedOperationException` in the unsupported functions.

sammcveety:
Added dummy class.  Please take another look.

dhalperi:
LGTM, unfortunately Checkstyle wins again

```
[INFO] Starting audit...
/Users/travis/build/GoogleCloudPlatform/DataflowJavaSDK/sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/TemplatingDataflowPipelineRunner.java:79:19: Name 'ERROR' must match pattern '^[a-z][a-zA-Z0-9]*_?$'.
/Users/travis/build/GoogleCloudPlatform/DataflowJavaSDK/sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/TemplatingDataflowPipelineRunner.java:81: Trailing whitespace
Audit done.
```

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/TemplatingDataflowPipelineRunner.java
```diff
@@ -0,0 +1,154 @@
...
+    DataflowPipelineRunner dataflowPipelineRunner =
+        DataflowPipelineRunner.fromOptions(dataflowOptions);
+    checkArgument(!Strings.isNullOrEmpty(dataflowOptions.getDataflowJobFile()),
+                  "--dataflowJobFile must be present to create a template.");
+
+    return new TemplatingDataflowPipelineRunner(dataflowPipelineRunner, options);
+  }
+
+  private static class TemplateJob extends DataflowPipelineJob {
+    static String ERROR =
```
dhalperi: needs to be `final` for Checkstyle to not complain

sammcveety:
Done!

PR #347, opened 2016-07-25
Adding PartitionedFileSink

NarekA:
This is a sink that expects KV objects as inputs and uses the key to partition the output contained in the values.


dhalperi:
Hi Narek,

Thanks for your contribution!

For the most part, new features for Dataflow are actually being added as part of the [Apache Beam (incubating)](https://github.com/apache/incubator-beam/) project, which will be the basis for Dataflow 2.0.0 release.

(Once accepted by Beam, critical features can be backported to Dataflow.)

It looks like you are partially implementing [BEAM-92](https://issues.apache.org/jira/browse/BEAM-92)?

Would you mind sending this PR to the Apache Beam? You can start with the [Contribution Guide](http://beam.incubator.apache.org/contribution-guide/).

Thanks,
Dan

PR #348, opened 2016-07-28
Try to find the active gcloud config

ravwojdyla:


dhalperi:
Hi Rafal,

Thanks -- this looks like a nice improvement.

It looks like it will apply equally well to Apache Beam (incubating), which will be the foundation for this SDK starting in version 2.0. We are try to add new features in Beam first, then backport them here -- this prevents future regressions.

Would you mind starting this PR there, and then we can backport it here? The [Beam Contribution Guide](http://beam.incubator.apache.org/contribution-guide/) is the place to get started.

Thanks!
Dan

dhalperi:
CLA confirmed for Spotify.

sdk/src/main/java/com/google/cloud/dataflow/sdk/options/GcpOptions.java
```diff
@@ -183,24 +184,67 @@ void setCredentialFactoryClass(
...
+      return System.getProperty("user.home");
+    }
+
+    private String getActiveConfig() throws IOException {
+      final String defaultConfigName = "default";
+
+      // get active configuration
+      File activeConfFile = new File(
+          getUserHome(),
+          ".config/gcloud/active_config");
```
lukecwik: Does this path work in windows?

sdk/src/main/java/com/google/cloud/dataflow/sdk/options/GcpOptions.java
```diff
@@ -183,24 +184,67 @@ void setCredentialFactoryClass(
...
+      if (config.size() != 1) {
+        LOG.debug("Active config file {} corrupted - use default",
+            activeConfFile.getCanonicalPath());
+        return defaultConfigName;
+      }
+
+      String configName = config.get(0);
+
+      if (configName.isEmpty()) {
+        LOG.debug("Active config in {} is empty - use default");
```
lukecwik: missing parameter for logging

ravwojdyla:
@dhalperi sure - will open PR in the Beam. Will finialize comments with @lukecwik here first tho.

sdk/src/main/java/com/google/cloud/dataflow/sdk/options/GcpOptions.java
ravwojdyla: Thanks - will fix.

sdk/src/main/java/com/google/cloud/dataflow/sdk/options/GcpOptions.java
ravwojdyla: Good question - there is `if` branch for `windows` before, also previous version of the code assumed the same hierarchy for all platforms ([here](https://github.com/GoogleCloudPlatform/DataflowJavaSDK/pull/348/files#diff-50c13c4ad4237983c6a869fe066c523dL196)). What would you suggest?

lukecwik: It seems as though Google is vending a library that already does this and instead of Dataflow owning this logic, we can rely on this library but wanted to make sure that this meets your needs:

https://github.com/GoogleCloudPlatform/gcloud-java/blob/master/gcloud-java-core/src/main/java/com/google/cloud/ServiceOptions.java
Look at the logic for defaultProject(), specifically activeGoogleCloudConfig() and googleCloudProjectId()

ravwojdyla: Sorry @lukecwik - was on vacation. I see, thanks for the gcloud-java link, I looked at the code, and it's pretty much the same. afaiu current DF does not check for:

```
    String projectId = System.getProperty(PROJECT_ENV_NAME, System.getenv(PROJECT_ENV_NAME));
    if (projectId == null) {
      projectId = appEngineProjectId();
    }
    if (projectId == null) {
      projectId = serviceAccountProjectId();
    }
```

Specifically `serviceAccountProjectId` could be interesting, afaiu it would discover if user is using a service account through `GOOGLE_APPLICATION_CREDENTIALS` and then if available, fetch the project from the credentials file. One could argue if that is good or bad idea, probably good for `defaultProject`.

It seems like we could push this PR, I could submit PR to gcloud-java to abstract the project discover logic away from `ServiceOptions`, then I could submit another PR to DF to start using gcloud-java? Does it sound like a good plan?

ravwojdyla: Ping @lukecwik

lukecwik:
Opened up https://github.com/GoogleCloudPlatform/google-cloud-java/issues/1207

sdk/src/main/java/com/google/cloud/dataflow/sdk/options/GcpOptions.java
ravwojdyla: ping 2x @lukecwik - do you think we should wait for google-cloud-java or move forward?

lukecwik: I would wait for google-cloud-java, it seems as though they are working towards the change as evidenced on GoogleCloudPlatform/google-cloud-java#1207

lukecwik:
This is superseded by the follow feature request on Apache Beam 2.x:
https://issues.apache.org/jira/browse/BEAM-2391

PR #354, opened 2016-08-04
Use Datastore AncestorKey for AutoComplete example

vikkyrk:
- To provide strongly consistent reads of entities written to Datastore, we need to use ancestor keys.


vikkyrk:
R: @dhalperi

PR #358, opened 2016-08-05
Use EqualsTester in ProxyInvocationHandlerTest

tgroh:
This tests the equality groups instead of manual assertions, including
same-reference equality.


tgroh:
R: @dhalperi

PR #361, opened 2016-08-12
Deduplicate Unbounded Reads

tgroh:
This ensures that sources that produce duplicate elements that are
marked as requiresDeduplication are handled by the DirectRunner.

Backports https://github.com/apache/incubator-beam/pull/473


tgroh:
R: @dhalperi

PR #363, opened 2016-08-12
Remove unneccessary method in ReadEvaluatorFactories

tgroh:
The getTransformEvaluator() method's only call is poll on the result of
getTransformEvaluatorQueue (which is only ever called by
getTransformEvaluator). Instead, move the construction to
getTransformEvaluator and call poll on the result.

Backports https://github.com/apache/incubator-beam/pull/747


tgroh:
R: @lukecwik

PR #364, opened 2016-08-12
Minor Cleanups in the Direct Runner

tgroh:
backports https://github.com/apache/incubator-beam/pull/795
backports https://github.com/apache/incubator-beam/pull/811


sdk/src/test/java/com/google/cloud/dataflow/sdk/runners/inprocess/InProcessPipelineRunnerTest.java
```diff
@@ -88,6 +91,50 @@ public String apply(KV<String, Long> input) {
     result.awaitCompletion();
   }
 
+  private static AtomicInteger changed;
```
lukecwik: Would have been better as a local final variable within the test method

tgroh: It can't be because if it is it gets cloned with the DoFn and we don't get the correct output.

PR #365, opened 2016-08-12
Make TimerData#compareTo consistent with equals

tgroh:
Timers are equal if the domain, timestamp, and namespace are equal.
Compare these values in compareTo. The ordering of TimerData that are
not in the same namespace or domain is arbitrary.

Update GroupAlsoByWindowsProperties to ignore ordering of output
values, as this is not part of the requirements of a GroupAlsoByWindows
implementation.

in part backports https://github.com/apache/incubator-beam/pull/813
Requires review for GroupAlsoByWindowsProperties update.

### DIFF
--- a/sdk/src/main/java/com/google/cloud/dataflow/sdk/util/TimerInternals.java
+++ b/sdk/src/main/java/com/google/cloud/dataflow/sdk/util/TimerInternals.java
@@ -28,6 +28,7 @@ import com.google.cloud.dataflow.sdk.util.state.StateNamespace;
 import com.google.cloud.dataflow.sdk.util.state.StateNamespaces;
 import com.google.common.base.MoreObjects;
 import com.google.common.base.Preconditions;
+import com.google.common.collect.ComparisonChain;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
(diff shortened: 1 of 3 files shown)


tgroh:
R: @bjchambers

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/TimerInternals.java
```diff
@@ -204,9 +205,21 @@ public String toString() {
...
+     * <p>The ordering of {@link TimerData} that are not in the same namespace or domain is
+     * arbitrary.
+     */
     @Override
     public int compareTo(TimerData o) {
-      return Long.compare(timestamp.getMillis(), o.getTimestamp().getMillis());
+      ComparisonChain chain = ComparisonChain.start()
+          .compare(timestamp.getMillis(), o.getTimestamp().getMillis())
+          .compare(domain, o.domain);
+      if (chain.result() == 0) {
```
bjchambers: I'm guessing this is being done because `namespace.stringKey()` may be expensive. If so comment on that, or else someone will suggest (or just clean this up) by suggesting getting rid of the `if`.

tgroh: Done.

PR #370, opened 2016-08-15
Exclude guava-testlib from shading relocation

swegner:
Previously, guava-testlib guava-testlib was being relocated as part of
the shading process, but test-scope dependencies aren't bundled in the
uber-jar. As a result, the output JAR was unusable without recreating
the same shading rules in a consuming project.

### DIFF
--- a/sdk/pom.xml
+++ b/sdk/pom.xml
@@ -360,6 +360,7 @@
                 <relocation>
                   <pattern>com.google.common</pattern>
                   <excludes>
+                    <!-- com.google.common is too generic, need to exclude guava-testlib -->
                     <exclude>com.google.common.**.testing</exclude>
                   </excludes>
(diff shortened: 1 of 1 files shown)


swegner:
R: @lukecwik

lukecwik:
This backports https://github.com/apache/incubator-beam/pull/832

lukecwik:
Add a comment stating com.google.common is too generic and we need to exclude a path that would match the guava-testlib dependency anywhere this exclude line is added.

Also, add a comment to everywhere the guava-testlib dependency is added (including root pom.xml inside dependencyManagement section) that if relocating com.google.common with the maven shade plugin, you should be warned that this will also match classes within guava-testlib and that you should exclude have an exclude for com.google.common.**.testing

lukecwik:
Minor comments:
when package relocating guava -> when relocating guava

and then LGTM

lukecwik:
LGTM, will merge after checks pass

PR #371, opened 2016-08-16
Add structured name support to dataflow batch worker

joshualitt:
This CL adds support for structured names and tests in the batch dataflow worker

### DIFF
--- a/sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/Counter.java
+++ b/sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/Counter.java
@@ -20,9 +20,12 @@ import static com.google.cloud.dataflow.sdk.util.common.Counter.AggregationKind.
 import static com.google.cloud.dataflow.sdk.util.common.Counter.AggregationKind.MEAN;
 import static com.google.cloud.dataflow.sdk.util.common.Counter.AggregationKind.OR;
 import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
 
+import com.google.auto.value.AutoValue;
 import com.google.cloud.dataflow.sdk.values.TypeDescriptor;
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.MoreObjects;
 import com.google.common.util.concurrent.AtomicDouble;
 
 import java.util.Objects;
@@ -114,6 +117,14 @@ public abstract class Counter<T> {
    * @throws IllegalArgumentException if the aggregation kind is not supported
    */
   public static Counter<Integer> ints(String name, AggregationKind kind) {
+    return new IntegerCounter(Name.withoutStructure(name), kind);
+  }
+
+  /**
+   * Like {@link #ints(String, AggregationKind)}, but takes a {@link Counter.Name} instead of a
+   * string.
+   */
+  public static Counter<Integer> ints(Counter.Name name, AggregationKind kind) {
     return new IntegerCounter(name, kind);
   }
 
@@ -129,6 +140,14 @@ public abstract class Counter<T> {
    * @throws IllegalArgumentException if the aggregation kind is not supported
    */
   public static Counter<Long> longs(String name, AggregationKind kind) {
+    return new LongCounter(Name.withoutStructure(name), kind);
+  }
+
+  /**
+   * Like {@link #longs(String, AggregationKind)}, but takes a {@link Counter.Name} instead of a
+   * string.
+   */
+  public static Counter<Long> longs(Counter.Name name, AggregationKind kind) {
     return new LongCounter(name, kind);
   }
 
@@ -144,6 +163,14 @@ public abstract class Counter<T> {
    * @throws IllegalArgumentException if the aggregation kind is not supported
    */
   public static Counter<Double> doubles(String name, AggregationKind kind) {
+    return new DoubleCounter(Name.withoutStructure(name), kind);
+  }
+
+  /**
+   * Like {@link #doubles(String, AggregationKind)}, but takes a {@link Counter.Name} instead of a
+   * string.
+   */
+  public static Counter<Double> doubles(Counter.Name name, AggregationKind kind) {
     return new DoubleCounter(name, kind);
   }
 
@@ -158,6 +185,14 @@ public abstract class Counter<T> {
    * @throws IllegalArgumentException if the aggregation kind is not supported
    */
   public static Counter<Boolean> booleans(String name, AggregationKind kind) {
+    return new BooleanCounter(Name.withoutStructure(name), kind);
+  }
+
+  /**
+   * Like {@link #booleans(String, AggregationKind)}, but takes a {@link Counter.Name} instead of a
+   * string.
+   */
+  public static Counter<Boolean> booleans(Counter.Name name, AggregationKind kind) {
     return new BooleanCounter(name, kind);
   }
 
@@ -173,9 +208,17 @@ public abstract class Counter<T> {
    */
   @SuppressWarnings("unused")
   private static Counter<String> strings(String name, AggregationKind kind) {
-    return new StringCounter(name, kind);
+    return new StringCounter(Name.withoutStructure(name), kind);
   }
 
+  /**
+   * Like {@link #strings(String, AggregationKind)}, but takes a {@link Counter.Name} instead of a
+   * string.
+   */
+  @SuppressWarnings("unused")
+  private static Counter<String> strings(Counter.Name name, AggregationKind kind) {
+    return new StringCounter(name, kind);
+  }
 
   //////////////////////////////////////////////////////////////////////////////
 
@@ -218,7 +261,12 @@ public abstract class Counter<T> {
    * Returns the counter's name.
    */
   public String getName() {
-    return name;
+    return this.name.counterName();
+  }
+
+  /** Returns a Name object which uniquely identifies this counter. */
+  public Name getUniqueName() {
+    return this.name;
   }
 
   /**
@@ -342,28 +390,26 @@ public abstract class Counter<T> {
    */
   @Override
   public String toString() {
-    StringBuilder sb = new StringBuilder();
-    sb.append(getName());
-    sb.append(":");
-    sb.append(getKind());
-    sb.append("(");
+    String valueString;
     switch (kind) {
       case SUM:
       case MAX:
       case MIN:
       case AND:
       case OR:
-        sb.append(getAggregate());
+        valueString = getAggregate().toString();
         break;
       case MEAN:
-        sb.append(getMean());
+        valueString = getMean().toString();
         break;
       default:
         throw illegalArgumentException();
     }
-    sb.append(")");
-
-    return sb.toString();
+    return MoreObjects.toStringHelper(this)
+      .add("Name", name)
+      .add("Kind", getKind())
+      .add("Value", valueString)
+      .toString();
   }
 
   @Override
@@ -418,17 +464,58 @@ public abstract class Counter<T> {
 
   //////////////////////////////////////////////////////////////////////////////
 
-  /** The name of this counter. */
-  protected final String name;
-
   /** The kind of aggregation function to apply to this counter. */
   protected final AggregationKind kind;
 
   /** The commit state of this counter. **/
   protected final AtomicReference<CommitState> commitState;
 
+  /**
+   * {@link Counter.Name} is the full unique name of the counter. Counters must have
+   * at least a name, and optionally may be associated with a {@code systemName}, an
+   * {@code originalName}, or both. All of these names together define a counter uniquely.
+   */
+  @AutoValue
+  public abstract static class Name {
+    /** Create a {@link Counter.Name} with the {@code originalName} from a {@link NameContext}. */
+    public static Name withOriginalName(String name, NameContext context) {
+      return new AutoValue_Counter_Name(name, null,
+          checkNotNull(context.originalName(), "Expected original name in context %s", context));
+    }
+
+    /** Create a {@link Counter.Name} with the {@code systemName} from a {@link NameContext}. */
+    public static Name withSystemName(String name, NameContext context) {
+      return new AutoValue_Counter_Name(name,
+          checkNotNull(context.systemName(), "Expected system name in context %s", context),
+          null);
+    }
+
+    /** Create a {@link Counter.Name} without a {@link NameContext}. */
+    public static Name withoutStructure(String name) {
+      return new AutoValue_Counter_Name(name, null, null);
+    }
+
+    /** Returns the non-unique {@code counterName}. */
+    public abstract String counterName();
+
+    /** Returns an optional {@code contextSystemName}. */
+    @Nullable public abstract String contextSystemName();
+
+    /** Returns an optional {@code contextOriginalName}. */
+    @Nullable public abstract String contextOriginalName();
+  }
+
+  /** The name of this counter. */
+  protected final Name name;
+
   protected Counter(String name, AggregationKind kind) {
-    this.name = name;
+    this.name = Name.withoutStructure(name);
+    this.kind = kind;
+    this.commitState = new AtomicReference<>(CommitState.COMMITTED);
+  }
+
+  protected Counter(Name name, AggregationKind kind) {
+    this.name = checkNotNull(name);
     this.kind = kind;
     this.commitState = new AtomicReference<>(CommitState.COMMITTED);
   }
@@ -445,7 +532,7 @@ public abstract class Counter<T> {
     private final AtomicReference<LongCounterMean> deltaMean;
 
     /** Initializes a new {@link Counter} for {@link Long} values. */
-    private LongCounter(String name, AggregationKind kind) {
+    private LongCounter(Counter.Name name, AggregationKind kind) {
       super(name, kind);
       switch (kind) {
         case MEAN:
@@ -657,7 +744,7 @@ public abstract class Counter<T> {
     AtomicReference<DoubleCounterMean> deltaMean;
 
     /** Initializes a new {@link Counter} for {@link Double} values. */
-    private DoubleCounter(String name, AggregationKind kind) {
+    private DoubleCounter(Counter.Name name, AggregationKind kind) {
       super(name, kind);
       switch (kind) {
         case MEAN:
@@ -867,7 +954,7 @@ public abstract class Counter<T> {
     private final AtomicBoolean deltaAggregate;
 
     /** Initializes a new {@link Counter} for {@link Boolean} values. */
-    private BooleanCounter(String name, AggregationKind kind) {
+    private BooleanCounter(Counter.Name name, AggregationKind kind) {
       super(name, kind);
       aggregate = new AtomicBoolean();
       deltaAggregate = new AtomicBoolean();
@@ -952,7 +1039,7 @@ public abstract class Counter<T> {
    */
   private static class StringCounter extends Counter<String> {
     /** Initializes a new {@link Counter} for {@link String} values. */
-    private StringCounter(String name, AggregationKind kind) {
+    private StringCounter(Counter.Name name, AggregationKind kind) {
       super(name, kind);
       // TODO: Support MIN, MAX of Strings.
       throw illegalArgumentException();
@@ -1036,7 +1123,7 @@ public abstract class Counter<T> {
     private final AtomicReference<IntegerCounterMean> deltaMean;
 
     /** Initializes a new {@link Counter} for {@link Integer} values. */
-    private IntegerCounter(String name, AggregationKind kind) {
+    private IntegerCounter(Counter.Name name, AggregationKind kind) {
       super(name, kind);
       switch (kind) {
         case MEAN:
--- a/sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/CounterSet.java
+++ b/sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/CounterSet.java
@@ -17,6 +17,7 @@
 package com.google.cloud.dataflow.sdk.util.common;
 
 import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
 
 import java.util.AbstractSet;
 import java.util.Iterator;
@@ -30,7 +31,7 @@ import java.util.concurrent.ConcurrentHashMap;
 public class CounterSet extends AbstractSet<Counter<?>> {
 
   /** Registered counters. */
-  private final ConcurrentHashMap<String, Counter<?>> counters = new ConcurrentHashMap<>();
+  private final ConcurrentHashMap<Counter.Name, Counter<?>> counters = new ConcurrentHashMap<>();
 
   private final AddCounterMutator addCounterMutator = new AddCounterMutator();
 
@@ -49,6 +50,15 @@ public class CounterSet extends AbstractSet<Counter<?>> {
    */
   public AddCounterMutator getAddCounterMutator() {
     return addCounterMutator;
+  }
+
+  /**
+   * Returns an object that supports adding additional counters into
+   * this {@link CounterSet} with the supplied {@link NameContext}.
+   */
+  public AddCounterMutator getAddCounterMutator(NameContext context) {
+    AddCounterMutator addCounterWithContextMutator = new AddCounterMutator(context);
+    return addCounterWithContextMutator;
   }
 
   /**
@@ -73,7 +83,7 @@ public class CounterSet extends AbstractSet<Counter<?>> {
    * name but an incompatible kind had already been added
    */
   public <T> Counter<T> addOrReuseCounter(Counter<T> counter) {
-    Counter<?> oldCounter = counters.putIfAbsent(counter.getName(), counter);
+    Counter<?> oldCounter = counters.putIfAbsent(counter.getUniqueName(), counter);
     if (oldCounter != null) {
       checkArgument(counter.isCompatibleWith(oldCounter),
         "Counter %s duplicates incompatible counter %s in %s", counter, oldCounter, this);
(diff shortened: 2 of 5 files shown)


joshualitt:
I signed it!

joshualitt:
I signed it!

joshualitt:
I signed it!

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/Counter.java
```diff
@@ -241,6 +246,35 @@ public AggregationKind getKind() {
    */
   public abstract T getAggregate();
 
+  /** Returns the naming context associated with this counter. */
+  public NameContext getNameContext() {
+    return this.name.getContext();
+  }
+
+  /** Sets the naming context associated with this counter. */
+  public void setNameContext(NameContext context, NameType type) {
```
bjchambers: Can you expand the Javadoc here?

`@Nullable` on context, and some explanation of the behavior? It has some weird requirements (if type == UNSTRUCTURED, and context != null then it throws an error).

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/Counter.java
```diff
@@ -241,6 +246,35 @@ public AggregationKind getKind() {
...
+    if (context != null) {
+      switch (type) {
+          case UNSTRUCTURED:
+            throw new IllegalArgumentException("Unstructured counters have no context");
```
bjchambers: I wonder if it would be better to have an "UnstructuredName(this.name.getName())" that we use here?

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/Counter.java
```diff
@@ -241,6 +246,35 @@ public AggregationKind getKind() {
...
+          case ORIGINAL:
+            this.name = new StructuredNameOriginal(this.name.getName(), context);
```
bjchambers: Should it instead throw an exception if we get no context along with type ORIGINAL/OPTIMIZED? Doesn't that represent a bug?

What if this instead just tried to create the StructuredNameOriginal and that had a Preconditions.checkNotNull that imposed the appropriate requirements?

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/Counter.java
```diff
@@ -418,17 +452,191 @@ public boolean isCompatibleWith(Counter<?> that) {
...
-  protected final String name;
-
   /** The kind of aggregation function to apply to this counter. */
   protected final AggregationKind kind;
 
   /** The commit state of this counter. **/
   protected final AtomicReference<CommitState> commitState;
 
+  /** which name to use from the original context. */
+  public static enum NameType {
```
bjchambers: Do we even need this? Can we just use one of the implementations of Name? (eg., have an UnstructuredName class, an OriginalName class and an OptimizedName class)?

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/Counter.java
```diff
@@ -418,17 +452,191 @@ public boolean isCompatibleWith(Counter<?> that) {
...
+    public String getName();
+    public NameContext getContext();
+    public NameType getType();
+  }
+
+  /** A StructuredName which uses the Original step name from the NameContext. */
+  public static class StructuredNameOriginal implements Name {
+
+    /** The name of this counter. */
+    private final String name;
```
bjchambers: This could use `@AutoValue` to implement most of the logic (eg., all the getters, the toString, the hashCode, the equals, etc.)

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/Counter.java
```diff
@@ -418,17 +452,191 @@ public boolean isCompatibleWith(Counter<?> that) {
...
+        StructuredNameOriginal that = (StructuredNameOriginal) other;
+        return this.name.equals(that.name)
+            && this.context.getOriginalName().equals(that.context.getOriginalName());
+      }
+      return false;
+    }
+  }
+
+  /** A StructuredName which uses the Optimized step name from the NameContext. */
+  public static class StructuredNameOptimized implements Name {
```
bjchambers: This could use `@AutoValue` to implement most of the logic (eg., all the getters, the toString, the hashCode, the equals, etc.)

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/NameContext.java
```diff
@@ -0,0 +1,70 @@
...
+ ******************************************************************************/
+package com.google.cloud.dataflow.sdk.util.common;
+
+import java.util.Objects;
+
+/**
+ * A NameContext provides a way for counters to control what context they will be associated with in
+ * monitoring.
+ */
+public class NameContext {
```
bjchambers: `@AutoValue`

sdk/src/test/java/com/google/cloud/dataflow/sdk/util/common/CounterSetTest.java
```diff
@@ -222,4 +226,110 @@ public void testAddCounterMutatorAddMultipleCounters() {
...
 
+  @Test
+  public void testHandlesDuplicateEntries() {
+    final String counterName = "counter-name";
+    CounterSet counters = new CounterSet();
+
+    // getting a nonexistent counter returns null
+    assertThat(counters.getExistingCounter(counterName), is(nullValue()));
+
+    // adding an already existing counter using addNewCounter throws an exception
```
bjchambers: Split these up into separate tests.
Use `@Rule public ExpectedException thrown = ExpectedException.none()` (as a field). Then:

```
@Test
public void testAddDuplicate() {
  Counter<Integer> counter = Counter.ints(counterName, SUM);
  counters.addNewCounter(counterAddCheck);
  thrown.expect(IllegalArgumentException.class);
  counters.addNewCounter(counterAddCheck);
}
```

sdk/src/test/java/com/google/cloud/dataflow/sdk/util/common/CounterSetTest.java
```diff
@@ -222,4 +226,110 @@ public void testAddCounterMutatorAddMultipleCounters() {
...
+      counters.addNewCounter(counterAddCheck);
+    } catch (IllegalArgumentException e) {
+      exceptionThrown = true;
+    }
+    assertThat(exceptionThrown, is(true));
+    assertThat(counters.getExistingCounter(counterName), is(notNullValue()));
+  }
+
+  @Test
+  public void testMutatorSanity() {
```
bjchambers: Better name and/or add a comment on what constitutes "sanity". Also, suggest splitting this up into more tests for specific cases.

sdk/src/test/java/com/google/cloud/dataflow/sdk/util/common/CounterSetTest.java
```diff
@@ -222,4 +226,110 @@ public void testAddCounterMutatorAddMultipleCounters() {
...
+    assertThat(counters.size(), is(size));
+    assertThat(counterUnstructuredToo.getAggregate(), is(10));
+
+    // This counter should still be unstructured, ie the two mutators should not interfere with one
+    // another
+    assertThat(counterUnstructuredToo.getNameType(), is(Counter.NameType.UNSTRUCTURED));
+  }
+
+  @Test
+  public void testEquivalentAdd() {
```
bjchambers: Update name to reflect what this is actually testing (and/or split this into multiple cases)

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/Counter.java
joshualitt: I'm not sure AutoValue adds much to these classes because they have to do custom hashCode / Equals logic to only compare parts of the name context which make sense for them. Thoughts?

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/Counter.java
```diff
@@ -418,17 +434,160 @@ public boolean isCompatibleWith(Counter<?> that) {
...
+    }
+
+    @Override
+    public NameContext getContext() {
+      return this.context;
+    }
+
+    @Override
+    public String toString() {
+      StringBuilder sb = new StringBuilder();
```
bjchambers: You can use:

```
return MoreObjects.toStringHelper(this)
    .add("Name", name)
    .add("Context", context)
    .toString();
```

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/Counter.java
```diff
@@ -418,17 +434,160 @@ public boolean isCompatibleWith(Counter<?> that) {
...
+    public int hashCode() {
+      return Objects.hash(name, context.originalName());
+    }
+
+    @Override
+    public boolean equals(Object other) {
+      if (other instanceof StructuredNameOriginal) {
+        StructuredNameOriginal that = (StructuredNameOriginal) other;
+        return this.name.equals(that.name)
+            && this.context.originalName().equals(that.context.originalName());
```
bjchambers: Now I'm a bit confused by this. If the context knows both its original name and optimized name, why would we ever not use the optimized name?

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/Counter.java
```diff
@@ -218,7 +220,17 @@
...
+  }
+
+  /** Returns a Name object which uniquely identifies this counter. */
+  public Name getUniqueName() {
+    return this.name;
+  }
+
+  /** Sets the full name of the counter. */
+  public void setName(Name counterName) {
+    this.name = checkNotNull(counterName);
```
bjchambers: Note that this `setName` operation is potentially risky. If the counter is in a map/set (which I think is the case in `CounterSet`) then this could change the `hashCode` / `equals` definition, which would lead to being unable to find the `Counter` in the map/set. Why is this mutation safe?

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/Counter.java
```diff
@@ -343,7 +355,7 @@ protected void setDirty() {
   @Override
   public String toString() {
     StringBuilder sb = new StringBuilder();
```
bjchambers: `MoreObjects.toStringHelper`

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/Counter.java
```diff
@@ -418,17 +430,44 @@ public boolean isCompatibleWith(Counter<?> that) {
...
+    }
+
+    public static StructuredName createWithSystemName(String name, NameContext context) {
+      return new AutoValue_Counter_StructuredName(name, checkNotNull(context.systemName()), null);
+    }
+  }
+
+  /** An UnstructuerdName is the default, ie no NameContext set. */
+  @AutoValue
+  public abstract static class UnstructuredName implements Name {
```
bjchambers: UnstructuredName = StructuredName right? Why not use the same thing, and just have a "createWithoutStructure(...)"?

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/NameContext.java
```diff
@@ -0,0 +1,33 @@
...
+ * A NameContext provides a way for counters to control what context they will be associated with in
+ * monitoring.
+ */
+@AutoValue
+public abstract class NameContext {
+  public static NameContext create(String originalName, String systemName) {
+    return new AutoValue_NameContext(originalName, systemName);
+  }
+
+  public abstract String systemName();
```
bjchambers: Javadoc on the public getters. What do they return? What does it mean?

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/Counter.java
```diff
@@ -418,17 +430,44 @@ public boolean isCompatibleWith(Counter<?> that) {
...
   /** The kind of aggregation function to apply to this counter. */
   protected final AggregationKind kind;
 
   /** The commit state of this counter. **/
   protected final AtomicReference<CommitState> commitState;
 
+  /** A hashable name structure which will uniquely identify a counter. */
+  public interface Name {
+    public String name();
+    @Nullable public String systemName();
```
bjchambers: I don't see us ever accessing the systemName/originalName of the counter. Could we do:

public interface Name {
  public String name();
}

public abstract static class StructuredName implemnets Name {
  abstract String contextSystemName();
  abstract String contextOriginalName();

  // static constructors
}
```

So that it isn't publicly exposed that there are different kinds of names?

Comment (bjchambers) [inline sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/Counter.java]: Calling these "name" is a bit odd. Could we call them "contextSystemName" or "contextNameSystem"?

File: sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/Counter.java [hunk 18]
```diff
@@ -418,17 +430,44 @@ public boolean isCompatibleWith(Counter<?> that) {
...
+    @Nullable public String originalName();
```
bjchambers: Javadoc describing these methods and why they are nulllable, etc.

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/Counter.java
```diff
@@ -114,6 +117,13 @@
    * @throws IllegalArgumentException if the aggregation kind is not supported
    */
   public static Counter<Integer> ints(String name, AggregationKind kind) {
+    return new IntegerCounter(Name.createWithoutStructure(name), kind);
```
bjchambers: Could be `Name.withoutStructure(...)` if you wanted something shorter, but `createWithoutStructure` is fine too.

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/Counter.java
```diff
@@ -129,6 +139,13 @@
    * @throws IllegalArgumentException if the aggregation kind is not supported
    */
   public static Counter<Long> longs(String name, AggregationKind kind) {
+    return new LongCounter(Name.createWithoutStructure(name), kind);
+  }
+
+  /**
+   * Like the above, only takes a {@link Counter.Name} instead of a string.
```
bjchambers: Nit: `Like {@link #longs(String, AggregationKind)} ...` rather than "Like the above"

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/Counter.java
```diff
@@ -144,6 +161,13 @@
    * @throws IllegalArgumentException if the aggregation kind is not supported
    */
   public static Counter<Double> doubles(String name, AggregationKind kind) {
+    return new DoubleCounter(Name.createWithoutStructure(name), kind);
+  }
+
+  /**
+   * Like the above, only takes a {@link Counter.Name} instead of a string.
```
bjchambers: Ditto.

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/Counter.java
```diff
@@ -158,6 +182,13 @@
    * @throws IllegalArgumentException if the aggregation kind is not supported
    */
   public static Counter<Boolean> booleans(String name, AggregationKind kind) {
+    return new BooleanCounter(Name.createWithoutStructure(name), kind);
+  }
+
+  /**
+   * Like the above, only takes a {@link Counter.Name} instead of a string.
```
bjchambers: Ditto.

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/Counter.java
```diff
@@ -342,28 +387,26 @@ protected void setDirty() {
...
-    sb.append("(");
+    String kindString;
     switch (kind) {
       case SUM:
       case MAX:
       case MIN:
       case AND:
       case OR:
-        sb.append(getAggregate());
+        kindString = getAggregate().toString();
```
bjchambers: `kindString` -> `valueString` or something like that?

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/Counter.java
```diff
@@ -418,17 +461,56 @@ public boolean isCompatibleWith(Counter<?> that) {
...
   protected final AtomicReference<CommitState> commitState;
 
+  /**
+   * The full unique name of the counter. Counters must have a
+   * at least a name, and optionally may be associated with
+   * a systemName, an originalName, or both. All of these
+   * names together define a counter uniquely.
+   */
+  @AutoValue
+  public abstract static class Name {
```
bjchambers: I'd consider calling this `CounterName` (1) to be closer to then names of classes in Beam and (2) since otherwise `Name` is relatively ambiguous. Fine to leave it as `Name` if the shorter-names are preferred.

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/Counter.java
```diff
@@ -418,17 +461,56 @@ public boolean isCompatibleWith(Counter<?> that) {
...
+    /** Create a Name with the originalName on a name context. */
+    public static Name createWithOriginalName(String name, NameContext context) {
+      return new AutoValue_Counter_Name(name, null, checkNotNull(context.originalName()));
```
bjchambers: Nit: `checkNotNull` should include messages so we can debug failures.

`checkNotNull(context.originalName(), "Expected original name in context %s", context)`

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/Counter.java
```diff
@@ -418,17 +461,56 @@ public boolean isCompatibleWith(Counter<?> that) {
...
+    }
+
+    /** Create a Name with the systemName on a name context. */
+    public static Name createWithSystemName(String name, NameContext context) {
+      return new AutoValue_Counter_Name(name, checkNotNull(context.systemName()), null);
```
bjchambers: Ditto

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/CounterSet.java
```diff
@@ -151,6 +170,26 @@ public void merge(CounterSet that) {
    * public methods of the CounterSet.
    */
   public class AddCounterMutator {
+
+    /**
+     * The NameContext to add counters into, optional.
```
bjchambers: `The {@link NameContext} to add ...`

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/CounterSet.java
```diff
@@ -164,5 +203,12 @@ public void merge(CounterSet that) {
     public <T> Counter<T> addCounter(Counter<T> counter) {
       return addOrReuseCounter(counter);
     }
+
+    /**
+     * Returns the NameContext associated with this mutator, if any.
```
bjchambers: `{@link NameContext}`

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/NameContext.java
```diff
@@ -0,0 +1,37 @@
...
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ ******************************************************************************/
+package com.google.cloud.dataflow.sdk.util.common;
+
+import com.google.auto.value.AutoValue;
+
+/**
+ * A NameContext provides a way for counters to control what context they will be associated with in
```
bjchambers: `The NameContext represents  the context in which a counter is reported. Specifically, it may be associated with a specific original step name and associated system name.`

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/NameContext.java
```diff
@@ -0,0 +1,37 @@
...
+ * monitoring.
+ */
+@AutoValue
+public abstract class NameContext {
+  /** Create a NameContext with both an originalName and systemName. */
```
bjchambers: `{@link NameContext}`, `{@code originalName}` etc.

Also -- what if we wanted counters that weren't associated with a step (eg., that was pipeline wide). Would that be a different kind of `NameContext`?

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/NameContext.java
```diff
@@ -0,0 +1,37 @@
...
+  public static NameContext create(String originalName, String systemName) {
+    return new AutoValue_NameContext(originalName, systemName);
+  }
+
+  /** Returns the systemName associated with this NameContext. */
```
bjchambers: Maybe useful to specify *what* constitutes a system name?

bjchambers: Nit: `{@link}` or `{@code}`

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/NameContext.java
```diff
@@ -0,0 +1,37 @@
...
+  public abstract String systemName();
+
+  /** Returns the originalName associated with this NameContext. */
```
bjchambers: Same nit

sdk/src/test/java/com/google/cloud/dataflow/sdk/util/common/CounterSetTest.java
```diff
@@ -222,4 +226,127 @@ public void testAddCounterMutatorAddMultipleCounters() {
     assertThat(set, containsInAnyOrder(c1, c2));
   }
 
+  @Test
+  public void testNullCounterBeforeAdd() {
+    final String counterName = "counter-name";
```
bjchambers: Can make this a static final constant in the class, so it doesn't need to be duplicated in all the tests.

sdk/src/test/java/com/google/cloud/dataflow/sdk/util/common/CounterSetTest.java
```diff
@@ -222,4 +226,127 @@ public void testAddCounterMutatorAddMultipleCounters() {
...
+    CounterSet counters = new CounterSet();
```
bjchambers: Can make this a final field in the test class, so it doesn't need to be created in each test.

sdk/src/test/java/com/google/cloud/dataflow/sdk/util/common/CounterSetTest.java
```diff
@@ -222,4 +226,127 @@ public void testAddCounterMutatorAddMultipleCounters() {
...
+    // adding an already existing counter using addNewCounter throws an exception
+    Counter<Integer> counterAddCheck = Counter.ints(counterName, SUM);
+    counters.addNewCounter(counterAddCheck);
+    thrown.expect(IllegalArgumentException.class);
+    counters.addNewCounter(counterAddCheck);
+    assertThat(counters.getExistingCounter(counterName), is(notNullValue()));
+  }
+
+  /**
+   * CounterSet mutators should not conflict with one another. The below tests
```
bjchambers: CounterSet mutators with different contexts should not conflict...

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/Counter.java
joshualitt: lets leave this as is, its so different from beam that I don't think its worth making changes to bring them closer together.

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/Counter.java
```diff
@@ -418,17 +464,58 @@ public boolean isCompatibleWith(Counter<?> that) {
...
   protected final AtomicReference<CommitState> commitState;
 
+  /**
+   * @{link Name} is the full unique name of the counter. Counters must have a
+   * at least a name, and optionally may be associated with a systemName, an originalName,
+   * or both. All of these names together define a counter uniquely.
+   */
+  @AutoValue
+  public abstract static class Name {
+    /** Create a {@link Name} with the originalName on a {@link NameContext}. */
```
bjchambers: "with the {@code originalName} from a ..."

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/Counter.java
```diff
@@ -418,17 +464,58 @@ public boolean isCompatibleWith(Counter<?> that) {
...
+    public static Name withOriginalName(String name, NameContext context) {
+      return new AutoValue_Counter_Name(name, null,
+          checkNotNull(context.originalName(), "Expected original name in context %s", context));
+    }
+
+    /** Create a {@link Name} with the systemName on a {@link NameContext}. */
```
bjchambers: `with the {@code systemName} from a ..."

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/Counter.java
```diff
@@ -418,17 +464,58 @@ public boolean isCompatibleWith(Counter<?> that) {
...
+    public static Name withSystemName(String name, NameContext context) {
+      return new AutoValue_Counter_Name(name,
+          checkNotNull(context.systemName(), "Expected system name in context %s", context),
+          null);
+    }
+
+    /** Create a {@link Name} with the originalName on a {@link NameContext}. */
```
bjchambers: `without a {@link NameContext}.`

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/common/Counter.java
```diff
@@ -418,17 +464,58 @@ public boolean isCompatibleWith(Counter<?> that) {
...
+    public static Name withoutStructure(String name) {
+      return new AutoValue_Counter_Name(name, null, null);
+    }
+
+    /** Returns the non-unique counter name. */
```
bjchambers: maybe `counterName()`?

And "Returns the counter name without any associated context."

bjchambers:
A few more doc comments and such, then I think this LGTM.

joshualitt:
Awesome, thanks for your patience with this review. Its important we get this right. Anyway, ptal

PR #374, opened 2016-08-16
Add configuration flag for Private IPs

sammcveety:
@dhalperi , could you please take a look?


sdk/src/main/java/com/google/cloud/dataflow/sdk/options/DataflowPipelineWorkerPoolOptions.java
```diff
@@ -254,4 +254,16 @@ public String getTeardownPolicyName() {
...
+  /**
+   * Specifies whether worker pools should be started with public IP addresses.
+   *
+   * <p>WARNING: This feature is experimental.  You must be whitelisted to use it.
+   */
+  @Description("Specifies whether worker pools should be started with public IP addresses. WARNING:"
+    + "This feature is experimental. You must be whitelisted to use it.")
+  @Experimental
+  @JsonIgnore
+  Boolean getUsePublicIps();
```
dhalperi: I notice that the option you send on the wire is a string. Do you anticipate future policies other than public/private? It just seems counter-intuitive to me that these are mismatched.

dhalperi: `@Nullable` on both getter and setter

dhalperi:
Minor questions, LGTM.

I was going to ask you to send this to apache/incubator-beam first, but then I realized this is a whitelist-only option and we since we don't support any version of Beam maybe that's unnecessary. For tracking sanity, I'd probably still prefer it go into Beam as well now so we don't forget later.

sammcveety:
Thanks!  Will send a PR for Beam once this is finalized.

Added Nullable.

For the bool vs. string, I'm relatively confident that bool is sufficient.  The reason it is tri-state internally is so that we can distinguish "unset" from an explicit toggle.

PR #376, opened 2016-08-17
Update CounterSet.java to be threadsafe

reuvenlax:
Some runners want to query CounterSet from a different thread than the one that mutates it. This requires thread safety.


dhalperi:
Thanks!

PR #377, opened 2016-08-17
Fix repackaging exclude pattern for guava-testlib

swegner:
The previous fix was incorrect in that the exclude pattern requires `.*` at the end. Tested this iteration using `javap` to inspect generated `.class` files.


swegner:
R: @lukecwik

PR #379, opened 2016-08-17
Make JAXBCoder Thread Safe

tgroh:
JAXB Marshaller and Unmarshaller are not thread safe, but coders are
required to be.

Create context only once, and create Marshaller and Unmarshaller locally
on-demand.

backports https://github.com/apache/incubator-beam/pull/497


lukecwik:
Using a transient ThreadLocal with an override on the initialValue would have been much cleaner:
https://docs.oracle.com/javase/7/docs/api/java/lang/ThreadLocal.html#initialValue()

initialValue is only called on first get() call per thread.

lukecwik:
LGTM

PR #384, opened 2016-08-18
Backport TriggerExampleTest fixes, Fix XMLSourceTest

tgroh:
TriggerExampleTest uses an equality method that does not match the assumptions
made by the test for qualities that make rows equal.

Use an explicit coder in XMLSourceTest to avoid having to make a change
to the underlying behavior of JAXBCoder that may be backwards incompatible
in some cases.


tgroh:
R: @bjchambers

sdk/src/test/java/com/google/cloud/dataflow/sdk/io/XmlSourceTest.java
```diff
@@ -227,6 +229,10 @@ public String toString() {
     }
   }
 
+
+  private Coder<Train> trainCoder =
+      AvroCoder.of(Train.class, ReflectData.AllowNull.get().getSchema(Train.class));
```
bjchambers: Comment on why we're using AvroCoder within an XmlSource test?

bjchambers:
The description of the PR doesn't seem accurate. This is changing more than just TriggerExampleTest.

tgroh:
Updated PR Description

sdk/src/test/java/com/google/cloud/dataflow/sdk/io/XmlSourceTest.java
tgroh: Done.

PR #393, opened 2016-08-22
Run rejected BigQueryTableInserter tasks in the calling thread

kennknowles:
Otherwise, even temporary bursts of data will cause failures as the thread pool's queue is saturated.

R: @dhalperi


dhalperi:
Per discussion with @dpmills , seems reasonable to block in the sink if this is the pipeline's bottleneck.

PR #394, opened 2016-08-23
Enable TextIO.Write in DataflowPipelineRunner for streaming

peihe:
This also back ports https://github.com/apache/incubator-beam/pull/802


peihe:
R: @dhalperi

peihe:
@dhalperi, why do we need to re-window the input to globalwindows?
If that is for performance, could we only re-window it for batch?

PR #395, opened 2016-08-23
Datastore Sink support for writing Mutations.

vikkyrk:
- Backport from https://github.com/apache/incubator-beam/pull/845
- SimpleFunction in DF SDK doesn't extend 'HasDisplayData', hence moving that to the PTransform. This is the only difference.


vikkyrk:
R: @dhalperi

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/datastore/V1Beta3.java
```diff
@@ -747,48 +859,40 @@ public void startBundle(Context c) {
...
+      mutations.add(c.element());
+      if (mutations.size() >= V1Beta3.DATASTORE_BATCH_UPDATE_LIMIT) {
         flushBatch();
       }
     }
 
     @Override
     public void finishBundle(Context c) throws Exception {
-      if (entities.size() > 0) {
+      if (mutations.size() > 0) {
```
dhalperi: missed this earlier -- this should just be `!mutations.isEmpty()`

Fix in followup.

dhalperi:
LGTM

PR #399, opened 2016-08-24
Fix TestPipeline javadoc to correspond with current implementation.

lukecwik:
R: @jasonkuster


dhalperi:
LGTM (jason is out or about to be, so I took over)

PR #400, opened 2016-08-24
Dynamically choose number of shards in the InProcessPipelineRunner

tgroh:
Add a Write Override Factory that limits the number of shards if
unspecified. This ensures that we will not write an output file per-key
due to bundling.
    
Do so by obtaining a count of the elements and obtaining the number of
shards based on the number of outputs.

backports https://github.com/apache/incubator-beam/pull/651


tgroh:
R: @dhalperi

PR #407, opened 2016-08-29
[BEAM-582] Allow usage of the new GCP service account JSON key

alexvanboxel:
This is a back port of a BEAM-582: With this PR you can use the newest JSON style service-account key files (over the now deprecated P12 key files). Note that the JSON case doesn't allow supplying an accountName, but the P12 requires it.

For more information look at:
https://github.com/apache/incubator-beam/pull/879
https://issues.apache.org/jira/browse/BEAM-582#


lukecwik:
This was superseded by Apache Beam 2.0.0 release and the following PR https://github.com/apache/beam/pull/1294

PR #408, opened 2016-08-29
Use Quiescence to Drive the DirectRunner

tgroh:
The Executor should add work whenever it becomes Quiescent.

Track the amount of outstanding work in the executor, and
modify the state appropriately whenever work is scheduled or completes.

Backports https://github.com/apache/incubator-beam/pull/754


tgroh:
R: @bjchambers

PR #410, opened 2016-08-29
Add the TestStream Primitive

tgroh:
This is a source suitable for use with tests that have interesting
triggering behavior. It is an Unbounded source that emits elements in
bundles, and advances the watermark and processing time appropriately.

TestStream is currently only supported in the InProcessPipelineRunner.

backports https://github.com/apache/incubator-beam/pull/817
backports https://github.com/apache/incubator-beam/pull/907

### DIFF
--- a/sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/InMemoryWatermarkManager.java
+++ b/sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/InMemoryWatermarkManager.java
@@ -39,6 +39,7 @@ import com.google.common.collect.Iterables;
 import com.google.common.collect.Ordering;
 import com.google.common.collect.SortedMultiset;
 import com.google.common.collect.TreeMultiset;
+import org.joda.time.Instant;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -55,7 +56,6 @@ import java.util.TreeSet;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.AtomicReference;
 import javax.annotation.Nullable;
-import org.joda.time.Instant;
 
 /**
  * Manages watermarks of {@link PCollection PCollections} and input and output watermarks of
--- a/sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/InMemoryWatermarkManager.java
+++ b/sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/InMemoryWatermarkManager.java
@@ -39,9 +39,6 @@ import com.google.common.collect.Iterables;
 import com.google.common.collect.Ordering;
 import com.google.common.collect.SortedMultiset;
 import com.google.common.collect.TreeMultiset;
-
-import org.joda.time.Instant;
-
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -57,8 +54,8 @@ import java.util.Set;
 import java.util.TreeSet;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.AtomicReference;
-
 import javax.annotation.Nullable;
+import org.joda.time.Instant;
 
 /**
  * Manages watermarks of {@link PCollection PCollections} and input and output watermarks of
@@ -1133,19 +1130,12 @@ class InMemoryWatermarkManager {
     private Map<StructuralKey<?>, FiredTimers> extractFiredTimers() {
       Map<StructuralKey<?>, List<TimerData>> eventTimeTimers =
           inputWatermark.extractFiredEventTimeTimers();
-      Map<StructuralKey<?>, List<TimerData>> processingTimers;
-      Map<StructuralKey<?>, List<TimerData>> synchronizedTimers;
-      if (inputWatermark.get().equals(BoundedWindow.TIMESTAMP_MAX_VALUE)) {
-        processingTimers = synchronizedProcessingInputWatermark.extractFiredDomainTimers(
-            TimeDomain.PROCESSING_TIME, BoundedWindow.TIMESTAMP_MAX_VALUE);
-        synchronizedTimers = synchronizedProcessingInputWatermark.extractFiredDomainTimers(
-            TimeDomain.PROCESSING_TIME, BoundedWindow.TIMESTAMP_MAX_VALUE);
-      } else {
-        processingTimers = synchronizedProcessingInputWatermark.extractFiredDomainTimers(
-            TimeDomain.PROCESSING_TIME, clock.now());
-        synchronizedTimers = synchronizedProcessingInputWatermark.extractFiredDomainTimers(
-            TimeDomain.SYNCHRONIZED_PROCESSING_TIME, getSynchronizedProcessingInputTime());
-      }
+      Map<StructuralKey<?>, List<TimerData>> processingTimers =
+          synchronizedProcessingInputWatermark.extractFiredDomainTimers(TimeDomain.PROCESSING_TIME,
+              clock.now());
+      Map<StructuralKey<?>, List<TimerData>> synchronizedTimers =
+          synchronizedProcessingInputWatermark.extractFiredDomainTimers(
+          TimeDomain.SYNCHRONIZED_PROCESSING_TIME, getSynchronizedProcessingInputTime());
       Map<StructuralKey<?>, Map<TimeDomain, List<TimerData>>> groupedTimers = new HashMap<>();
       groupFiredTimers(groupedTimers, eventTimeTimers, processingTimers, synchronizedTimers);
 
--- a/sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/InProcessPipelineOptions.java
+++ b/sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/InProcessPipelineOptions.java
@@ -15,17 +15,13 @@
  */
 package com.google.cloud.dataflow.sdk.runners.inprocess;
 
-import com.google.cloud.dataflow.sdk.Pipeline;
 import com.google.cloud.dataflow.sdk.options.ApplicationNameOptions;
 import com.google.cloud.dataflow.sdk.options.Default;
 import com.google.cloud.dataflow.sdk.options.Description;
 import com.google.cloud.dataflow.sdk.options.Hidden;
 import com.google.cloud.dataflow.sdk.options.PipelineOptions;
-import com.google.cloud.dataflow.sdk.options.Validation.Required;
 import com.google.cloud.dataflow.sdk.transforms.PTransform;
-
 import com.fasterxml.jackson.annotation.JsonIgnore;
-
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
 
@@ -37,17 +33,12 @@ public interface InProcessPipelineOptions extends PipelineOptions, ApplicationNa
    * Gets the {@link ExecutorServiceFactory} to use to create instances of {@link ExecutorService}
    * to execute {@link PTransform PTransforms}.
    *
-   * <p>Note that {@link ExecutorService ExecutorServices} returned by the factory must ensure that
-   * it cannot enter a state in which it will not schedule additional pending work unless currently
-   * scheduled work completes, as this may cause the {@link Pipeline} to cease processing.
-   *
    * <p>Defaults to a {@link FixedThreadPoolExecutorServiceFactory}, which produces instances of
    * {@link Executors#newCachedThreadPool()}.
    */
+  @Deprecated
   @JsonIgnore
-  @Required
   @Hidden
-  @Default.InstanceFactory(FixedThreadPoolExecutorServiceFactory.class)
   ExecutorServiceFactory getExecutorServiceFactory();
 
   void setExecutorServiceFactory(ExecutorServiceFactory executorService);
@@ -56,9 +47,8 @@ public interface InProcessPipelineOptions extends PipelineOptions, ApplicationNa
    * Gets the {@link Clock} used by this pipeline. The clock is used in place of accessing the
    * system time when time values are required by the evaluator.
    */
-  @Default.InstanceFactory(NanosOffsetClock.Factory.class)
+  @Deprecated
   @JsonIgnore
-  @Required
   @Hidden
   @Description(
       "The processing time source used by the pipeline. When the current time is "
--- a/sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/InProcessPipelineOptions.java
+++ b/sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/InProcessPipelineOptions.java
@@ -35,6 +35,8 @@ public interface InProcessPipelineOptions extends PipelineOptions, ApplicationNa
    *
    * <p>Defaults to a {@link FixedThreadPoolExecutorServiceFactory}, which produces instances of
    * {@link Executors#newCachedThreadPool()}.
+   *
+   * @deprecated the runner manages its own {@link ExecutorService} as an implementation detail
    */
   @Deprecated
   @JsonIgnore
@@ -46,6 +48,8 @@ public interface InProcessPipelineOptions extends PipelineOptions, ApplicationNa
   /**
    * Gets the {@link Clock} used by this pipeline. The clock is used in place of accessing the
    * system time when time values are required by the evaluator.
+   *
+   * @deprecated the runner manages its own {@link Clock} as an implementation detail
    */
   @Deprecated
   @JsonIgnore
--- a/sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/InProcessPipelineRunner.java
+++ b/sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/InProcessPipelineRunner.java
@@ -28,7 +28,9 @@ import com.google.cloud.dataflow.sdk.runners.PipelineRunner;
 import com.google.cloud.dataflow.sdk.runners.inprocess.GroupByKeyEvaluatorFactory.InProcessGroupByKeyOnly;
 import com.google.cloud.dataflow.sdk.runners.inprocess.GroupByKeyEvaluatorFactory.InProcessGroupByKeyOverrideFactory;
 import com.google.cloud.dataflow.sdk.runners.inprocess.InProcessCreate.InProcessCreateOverrideFactory;
+import com.google.cloud.dataflow.sdk.runners.inprocess.TestStreamEvaluatorFactory.InProcessTestStreamFactory;
 import com.google.cloud.dataflow.sdk.runners.inprocess.ViewEvaluatorFactory.InProcessViewOverrideFactory;
+import com.google.cloud.dataflow.sdk.testing.TestStream;
 import com.google.cloud.dataflow.sdk.transforms.Aggregator;
 import com.google.cloud.dataflow.sdk.transforms.AppliedPTransform;
 import com.google.cloud.dataflow.sdk.transforms.Create;
@@ -48,6 +50,8 @@ import com.google.cloud.dataflow.sdk.values.PCollectionView;
 import com.google.cloud.dataflow.sdk.values.PInput;
 import com.google.cloud.dataflow.sdk.values.POutput;
 import com.google.cloud.dataflow.sdk.values.PValue;
+import com.google.common.base.Supplier;
+import com.google.common.base.Suppliers;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
@@ -56,6 +60,7 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 
 /**
  * An In-Memory implementation of the Dataflow Programming Model. Supports Unbounded
@@ -77,6 +82,7 @@ public class InProcessPipelineRunner
               .put(Create.Values.class, new InProcessCreateOverrideFactory())
               .put(GroupByKey.class, new InProcessGroupByKeyOverrideFactory())
               .put(CreatePCollectionView.class, new InProcessViewOverrideFactory())
+              .put(TestStream.class, new InProcessTestStreamFactory())
               .put(Write.Bound.class, new WriteWithShardingFactory())
               .build();
 
@@ -173,6 +179,8 @@ public class InProcessPipelineRunner
 
   ////////////////////////////////////////////////////////////////////////////////////////////////
   private final InProcessPipelineOptions options;
+  private Supplier<ExecutorService> executorServiceSupplier;
+  private Supplier<Clock> clockSupplier;
 
   public static InProcessPipelineRunner fromOptions(PipelineOptions options) {
     return new InProcessPipelineRunner(options.as(InProcessPipelineOptions.class));
@@ -180,6 +188,17 @@ public class InProcessPipelineRunner
 
   private InProcessPipelineRunner(InProcessPipelineOptions options) {
     this.options = options;
+    if (options.getExecutorServiceFactory() == null) {
+      executorServiceSupplier = new FixedThreadPoolSupplier();
+    } else {
+      executorServiceSupplier =
+          new ExecutorServiceFactorySupplier(options.getExecutorServiceFactory());
+    }
+    if (options.getClock() == null) {
+      clockSupplier = new NanosOffsetClockSupplier();
+    } else {
+      clockSupplier = Suppliers.ofInstance(options.getClock());
+    }
   }
 
   /**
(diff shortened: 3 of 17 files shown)


tgroh:
R: @bjchambers

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/InProcessPipelineOptions.java
```diff
@@ -67,7 +56,7 @@
 
   void setClock(Clock clock);
 
-  @Default.Boolean(false)
+  @Default.Boolean(true)
```
bjchambers: Seems like a semantic change for users who weren't setting this -- why is this OK?

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/InProcessPipelineOptions.java
```diff
@@ -56,9 +47,7 @@
    * Gets the {@link Clock} used by this pipeline. The clock is used in place of accessing the
    * system time when time values are required by the evaluator.
    */
-  @Default.InstanceFactory(NanosOffsetClock.Factory.class)
   @JsonIgnore
-  @Required
```
bjchambers: Why is this no longer required? Should it be deprecated?

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/InProcessPipelineRunner.java
```diff
@@ -173,13 +179,26 @@
...
+  private Supplier<Clock> clockSupplier;
 
   public static InProcessPipelineRunner fromOptions(PipelineOptions options) {
     return new InProcessPipelineRunner(options.as(InProcessPipelineOptions.class));
   }
 
   private InProcessPipelineRunner(InProcessPipelineOptions options) {
     this.options = options;
+    if (options.getExecutorServiceFactory() == null) {
+      executorServiceSupplier = new FixedThreadPoolSupplier();
```
bjchambers: Why not use FixedThreadPoolSupplier as the default?

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/InProcessPipelineRunner.java
```diff
@@ -173,13 +179,26 @@
...
+    } else {
+      executorServiceSupplier =
+          new ExecutorServiceFactorySupplier(options.getExecutorServiceFactory());
+    }
+    if (options.getClock() == null) {
```
bjchambers: Why not leave NanosOffsetClock as the default in the options?

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/InProcessPipelineRunner.java
```diff
@@ -189,6 +208,14 @@ public InProcessPipelineOptions getPipelineOptions() {
     return options;
   }
 
+  Supplier<Clock> getClockSupplier() {
+    return clockSupplier;
+  }
+
+  void setClockSupplier(Supplier<Clock> supplier) {
```
bjchambers: I worry about changes in behavior when the clock is changed midway through processing. Why does it need to be mutable?

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/NanosOffsetClock.java
```diff
@@ -44,15 +40,5 @@ public Instant now() {
         baseMillis + (TimeUnit.MILLISECONDS.convert(
             System.nanoTime() - nanosAtBaseMillis, TimeUnit.NANOSECONDS)));
   }
-
-  /**
-   * Creates instances of {@link NanosOffsetClock}.
-   */
-  public static class Factory implements DefaultValueFactory<Clock> {
```
bjchambers: This change seems to be noise -- why did this need to change for the TestStream changes?

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/TestStreamEvaluatorFactory.java
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  private <InputT, OutputT> TransformEvaluator<? super InputT> createEvaluator(
+      AppliedPTransform<PBegin, PCollection<OutputT>, TestStream<OutputT>> application,
+      InProcessEvaluationContext evaluationContext) {
+    if (evaluator.get() == null) {
+      Evaluator<OutputT> createdEvaluator = new Evaluator<>(application, evaluationContext, inUse);
+      evaluator.compareAndSet(null, createdEvaluator);
+    }
+    if (inUse.compareAndSet(false, true)) {
```
bjchambers: This logic is somewhat confusing -- can you add javadoc on why you need to do this? At face value, creating an evaluator should be as simple as "return new Evaluator<>(application, evaluationContext, inUse)". I then see that you're caching the evaluator. That makes sense. But I'm not sure why you return null if you've already created one.

(1) Just make which ever thread successfully creates the evaluator be the one to return it?
(2) Or just have a boolean, and only the thread that flips false->true returns the evaluator.

(And also comment why you want to return after the first one)

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/TestStreamEvaluatorFactory.java
```diff
@@ -0,0 +1,195 @@
...
+import java.util.List;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+import javax.annotation.Nullable;
+
+/**
+ * The {@link TransformEvaluatorFactory} for the {@link TestStream} primitive.
+ */
+class TestStreamEvaluatorFactory implements TransformEvaluatorFactory {
+  private final AtomicBoolean inUse = new AtomicBoolean(false);
```
bjchambers: Javadoc on the purpose of these fields (especially how they should be used safely)?

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/TestStreamEvaluatorFactory.java
```diff
@@ -0,0 +1,195 @@
...
+  }
+
+  private static class TestClockSupplier implements Supplier<Clock> {
+    @Override
+    public Clock get() {
+      return new TestClock();
+    }
+  }
+
+  static class DirectTestStreamFactory implements PTransformOverrideFactory {
```
bjchambers: Can the definition of the test version exist in testing code somewhere?

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/TestStreamEvaluatorFactory.java
```diff
@@ -0,0 +1,195 @@
...
+
+      @Override
+      public PCollection<T> apply(PBegin input) {
+        setup(input.getPipeline());
+        return PCollection.<T>createPrimitiveOutputInternal(
+                input.getPipeline(), WindowingStrategy.globalDefault(), IsBounded.UNBOUNDED)
+            .setCoder(original.getValueCoder());
+      }
+
+      private void setup(Pipeline p) {
```
bjchambers: Not entirely clear why this had to be broken out -- were you intending other places to call setup?

Seems like this does both some validation and some mutation. It seems odd that having a DirectTestStream mutates the clock supplier for the whole pipeline. What if you had both a DirectTestStream and other non-DirectTestStream sources?

sdk/src/main/java/com/google/cloud/dataflow/sdk/testing/DataflowAssert.java
```diff
@@ -731,6 +748,12 @@ private PCollectionViewIterableAssert(
     }
 
     @Override
+    public IterableAssert<T> inEarlyGlobalWindowPanes() {
+      throw new UnsupportedOperationException(
+          "IterableAssert of an PCollectionView does not support windowed assertions");
```
bjchambers: "of a PCollectionView"

sdk/src/main/java/com/google/cloud/dataflow/sdk/testing/TestStream.java
```diff
@@ -0,0 +1,366 @@
...
+import java.io.OutputStream;
+import java.util.Collections;
+import java.util.List;
+
+/**
+ * A testing input that generates an unbounded {@link PCollection} of elements, advancing the
+ * watermark and processing time as elements are emitted. After all of the specified elements are
+ * emitted, ceases to produce output.
+ *
+ * <p>Each call to a {@link TestStream.Builder} method will only be reflected in the state of the
```
bjchambers: This Javadoc should be replicated on the builder and/or associated methods.

sdk/src/main/java/com/google/cloud/dataflow/sdk/testing/TestStream.java
```diff
@@ -0,0 +1,366 @@
...
+     */
+    @SafeVarargs
+    public final Builder<T> addElements(
+        TimestampedValue<T> element, TimestampedValue<T>... elements) {
+      checkArgument(
+          element.getTimestamp().isBefore(BoundedWindow.TIMESTAMP_MAX_VALUE),
+          "Elements must have timestamps before %s. Got: %s",
+          BoundedWindow.TIMESTAMP_MAX_VALUE,
+          element.getTimestamp());
+      for (TimestampedValue<T> multiElement : elements) {
```
bjchambers: the name "multiElement" makes this confusing, because I thought looking in the code that it was something special, not just one of the elements). I'd suggest renamining the parameters to "firstElement" and "elements" then having the loop be "for element in elements"

sdk/src/main/java/com/google/cloud/dataflow/sdk/testing/TestStream.java
```diff
@@ -0,0 +1,366 @@
...
+  /**
+   * The types of {@link Event} that are supported by {@link TestStream}.
+   */
+  public enum EventType {
+    ELEMENT,
+    WATERMARK,
+    PROCESSING_TIME
+  }
+
+  /** A {@link Event} that produces elements. */
```
bjchambers: An Event

sdk/src/main/java/com/google/cloud/dataflow/sdk/testing/TestStream.java
```diff
@@ -0,0 +1,366 @@
...
+    static <T> Event<T> advanceBy(Duration amount) {
+      return new AutoValue_TestStream_ProcessingTimeEvent<>(EventType.PROCESSING_TIME, amount);
+    }
+  }
+
+  @Override
+  public PCollection<T> apply(PBegin input) {
+    throw new IllegalStateException(
+        String.format(
+            "Pipeline Runner %s does not provide a required override for %s",
```
bjchambers: "required override" seems ambiguous (could mean `@Override` method). Maybe just "does not support %s" "does not provide/support an implementation of %s"?

sdk/src/test/java/com/google/cloud/dataflow/sdk/testing/TestStreamTest.java
```diff
@@ -0,0 +1,328 @@
...
+        VarIntCoder.of())
+        .addElements(TimestampedValue.of(1, instant),
+            TimestampedValue.of(2, instant),
+            TimestampedValue.of(3, instant))
+        .advanceWatermarkTo(instant.plus(Duration.standardMinutes(6)))
+        // These elements are late but within the allowed lateness
+        .addElements(TimestampedValue.of(4, instant), TimestampedValue.of(5, instant))
+        .advanceWatermarkTo(instant.plus(Duration.standardMinutes(20)))
+        // These elements are droppably late
+        .addElements(TimestampedValue.of(-1, instant),
```
bjchambers: Test would likely be easier to debug if it was separated into different cases -- eg., simplest case, dropping case (just verifying that data was dropped), triggering case, etc.

sdk/src/test/java/com/google/cloud/dataflow/sdk/testing/TestStreamTest.java
```diff
@@ -0,0 +1,328 @@
...
+    p.run();
+  }
+
+  @Test
+  public void testProcessingTimeTrigger() {
+    TestStream<Long> source = TestStream.create(
+        VarLongCoder.of())
+        .addElements(TimestampedValue.of(1L, new Instant(1000L)),
+            TimestampedValue.of(2L, new Instant(2000L)))
+        .advanceProcessingTime(Duration.standardMinutes(12))
```
bjchambers: I liked that in the watermark case we're using explicit times rather than durations. Should we instead have `advanceProcessingTimeTo(Instant)` to keep the two consistent (and make testing things like alignedTo and such easier)?

sdk/src/test/java/com/google/cloud/dataflow/sdk/testing/TestStreamTest.java
```diff
@@ -0,0 +1,328 @@
...
+            .advanceWatermarkTo(new Instant(1001L))
+            .addElements(
+                TimestampedValue.of("finalLatePane", new Instant(750)),
+                TimestampedValue.of("alsoFinalLatePane", new Instant(250)))
+            .advanceWatermarkToInfinity();
+
+    Pipeline p = getPipeline();
+    FixedWindows windowFn = FixedWindows.of(Duration.millis(1000L));
+    Duration allowedLateness = Duration.millis(5000L);
+    PCollection<String> values =
```
bjchambers: A comment on why the given pipeline actually tests discarding mode. Specifically, here we're testing discarding mode because we trigger multiple times, and if we were in accumulating mode, the elements in the first trigger firing would be duplicated.

sdk/src/test/java/com/google/cloud/dataflow/sdk/testing/TestStreamTest.java
```diff
@@ -0,0 +1,328 @@
...
+  public void testElementAtPositiveInfinityThrows() {
+    Builder<Integer> stream =
+        TestStream.create(VarIntCoder.of())
+            .addElements(TimestampedValue.of(-1, BoundedWindow.TIMESTAMP_MAX_VALUE.minus(1L)));
+    thrown.expect(IllegalArgumentException.class);
+    stream.addElements(TimestampedValue.of(1, BoundedWindow.TIMESTAMP_MAX_VALUE));
+  }
+
+  @Test
+  public void testAdvanceWatermarkNonMonotonicThrows() {
```
bjchambers: Add a test that you can start with negative watermarks?

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/InProcessPipelineOptions.java
tgroh: Reverted, but set in the TestStream override

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/InProcessPipelineOptions.java
tgroh: Yes.

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/InProcessPipelineRunner.java
tgroh: I have to set it somewhere during construction; it seems reasonable to only create the supplier that will be used.

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/InProcessPipelineRunner.java
tgroh: The runner will obtain an instance of the Clock during the call to `run()` and pass it to the `EvaluationContext`; so each is independent.

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/NanosOffsetClock.java
tgroh: I needed to replace the way clocks are obtained so multiple pipelines don't interact with each other if they're both using `TestStream`; that requires suppliers, which means we can have a default supplier instead of a DefaultValueFactory.

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/TestStreamEvaluatorFactory.java
tgroh: Race condition averted. This should be keyed by application.

Evaluators are stateful (so we can produce the correct outputs) so they have to be cached, but can't be split or parallelized otherwise we get output faster than is expected.

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/TestStreamEvaluatorFactory.java
tgroh: Doc'd the one remaining field

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/TestStreamEvaluatorFactory.java
tgroh: I don't understand what this means. This is the runner-specific override.

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/TestStreamEvaluatorFactory.java
tgroh: inlined.

sdk/src/main/java/com/google/cloud/dataflow/sdk/testing/DataflowAssert.java
tgroh: Fixed everywhere.

sdk/src/main/java/com/google/cloud/dataflow/sdk/testing/TestStream.java
tgroh: Isn't it? Every builder method has "like this one that will (do a thing) after all earlier events have completed"

sdk/src/main/java/com/google/cloud/dataflow/sdk/testing/TestStream.java
tgroh: first & rest
for element in rest

sdk/src/main/java/com/google/cloud/dataflow/sdk/testing/TestStream.java
tgroh: Done.

sdk/src/main/java/com/google/cloud/dataflow/sdk/testing/TestStream.java
tgroh: "does not support" seems fine

sdk/src/test/java/com/google/cloud/dataflow/sdk/testing/TestStreamTest.java
tgroh: Elements are output with an absolute timestamp, and the watermark is relative to these timestamps. However, Processing Time is always relative to whenever the pipeline happens to be executing, so users should not think of absolute processing times.

If in the future we decide it's important to have both absolute and relative references to processing time, we can revisit this.

sdk/src/test/java/com/google/cloud/dataflow/sdk/testing/TestStreamTest.java
tgroh: Done.

sdk/src/test/java/com/google/cloud/dataflow/sdk/testing/TestStreamTest.java
tgroh: Updated this one.

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/InProcessPipelineRunner.java
tgroh: Symmetry, and the removal in Beam made this an easier backport

tgroh:
Merged in changes from incubator head, namely https://github.com/apache/incubator-beam/pull/907

PTAL

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/InMemoryWatermarkManager.java
```diff
@@ -1133,19 +1130,12 @@ private void addPending(CommittedBundle<?> bundle) {
...
-        processingTimers = synchronizedProcessingInputWatermark.extractFiredDomainTimers(
-            TimeDomain.PROCESSING_TIME, clock.now());
-        synchronizedTimers = synchronizedProcessingInputWatermark.extractFiredDomainTimers(
-            TimeDomain.SYNCHRONIZED_PROCESSING_TIME, getSynchronizedProcessingInputTime());
-      }
+      Map<StructuralKey<?>, List<TimerData>> processingTimers =
+          synchronizedProcessingInputWatermark.extractFiredDomainTimers(TimeDomain.PROCESSING_TIME,
+              clock.now());
+      Map<StructuralKey<?>, List<TimerData>> synchronizedTimers =
+          synchronizedProcessingInputWatermark.extractFiredDomainTimers(
```
bjchambers: It seems weird that both of these methods are on `synchronizedProcessingInputWatermark`. I see the `TimeDomain` parameter being used to disambiguate, which I guess makes sense. Just wanted to confirm.

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/InProcessPipelineOptions.java
```diff
@@ -56,9 +47,8 @@
    * Gets the {@link Clock} used by this pipeline. The clock is used in place of accessing the
    * system time when time values are required by the evaluator.
    */
-  @Default.InstanceFactory(NanosOffsetClock.Factory.class)
+  @Deprecated
```
bjchambers: Add Javadoc on why this is deprecated/what to do instead.

sdk/src/main/java/com/google/cloud/dataflow/sdk/testing/TestStream.java
```diff
@@ -0,0 +1,366 @@
...
+      for (int i = 0; i < rest.length; i++) {
+        remainingElements[i] = TimestampedValue.of(rest[i], currentWatermark);
+      }
+      return addElements(firstElement, remainingElements);
+    }
+
+    /**
+     * Adds the specified elements to the source with the provided timestamps.
+     *
+     * @return A {@link TestStream.Builder} like this one that will add the provided elements
```
bjchambers: Should "add" be "emit"? The method is adding elements, the stream will cause those elements to be emitted after all earlier events have completed

sdk/src/main/java/com/google/cloud/dataflow/sdk/testing/TestStream.java
```diff
@@ -0,0 +1,366 @@
...
+    private Builder(Coder<T> coder, ImmutableList<Event<T>> events, Instant currentWatermark) {
+      this.coder = coder;
+      this.events = events;
+      this.currentWatermark = currentWatermark;
+    }
+
+    /**
+     * Adds the specified elements to the source with timestamp equal to the current watermark.
+     *
+     * @return A {@link TestStream.Builder} like this one that will add the provided elements
```
bjchambers: add -> emit?

bjchambers:
Nits then LGTM

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/InMemoryWatermarkManager.java
tgroh: Yes. Processing Time proper doesn't have a meaningful watermark, but processing time timers impact the SynchronizedProcessingTimeWatermark, so they're tracked together.

sdk/src/main/java/com/google/cloud/dataflow/sdk/testing/TestStream.java
tgroh: Sure.

sdk/src/main/java/com/google/cloud/dataflow/sdk/testing/TestStream.java
tgroh: Sure.

sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/inprocess/InProcessPipelineOptions.java
tgroh: Done.

PR #413, opened 2016-09-02
Add LeaderBoardTest

tgroh:
This test exercises the PTransforms that make up the LeaderBoard
example. This includes speculative and late trigger firings to produce
team and individual scores on a global and fixed window basis.

Refactor LeaderBoard to expose the team and user score calculations as
composite PTransforms to enable this testing.

backports https://github.com/apache/incubator-beam/pull/854


tgroh:
R: @lukecwik

lukecwik:
Diff'd against Apache Beam version, other than changes for DataflowExampleUtils, diff was the same.

PR #418, opened 2016-09-06
BigQuery: limit max job polling time to 1 minute

dhalperi:
Before the backoff would grow unboundedly, so we could in principle wait
1.5x to 2x the actual job time. For long running jobs this is hours.
Now, we just back off at most 1 minute between checking the job state.
Note there should be no danger of QPS overload here because we should
have very few concurrent outstanding jobs

This is a backport of apache/incubator-beam#917


dhalperi:
R: @tgroh

tgroh:
LGTM

PR #411, opened 2016-08-29
Create the temp dataset in the same location as the queried table in BigQueryTableRowIterator

peihe:
This fixed https://github.com/GoogleCloudPlatform/DataflowJavaSDK/issues/405

PR #420, opened 2016-09-07
BigQueryTableRowIterator: support queries that refernece no tables

dhalperi:
The new code in https://github.com/GoogleCloudPlatform/DataflowJavaSDK/pull/411 did
not adequately handle the case in which a query does not reference any tables, which
could result in a NullPointerException.


dhalperi:
R: @peihe (@lukecwik OR @bjchambers)

I'm still trying to discern if Pei is on vacation, so have backoff reviewers :).

peihe:
LGTM

Does Beam need this fix?

dhalperi:
You didn't port #411 to beam yet, but yes.

PR #430, opened 2016-09-13
checkstyle: prohibit API client repackaged Guava

dhalperi:
Apparently the IllegalImport check only blocks packages,
so we had to move to Regexp to get individual classes.

This is a backport of apache/incubator-beam#872


dhalperi:
R: @bjchambers

PR #431, opened 2016-09-13
BigQueryIO: enable users to control the SQL dialect

dhalperi:
BigQuery 2.0 adds a new dialect of SQL, Standard SQL, and we need to
enable passing a specific option to BigQuery's APIs in order to support
it.

### DIFF
--- a/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BigQueryIO.java
+++ b/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BigQueryIO.java
@@ -433,7 +433,8 @@ public class BigQueryIO {
        */
       public Bound named(String name) {
         return new Bound(
-            name, query, jsonTableRef, validate, flattenResults, useLegacySql, testBigQueryServices);
+            name, query, jsonTableRef, validate, flattenResults, useLegacySql,
+            testBigQueryServices);
       }
 
       /**
--- a/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BigQueryIO.java
+++ b/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BigQueryIO.java
@@ -520,69 +520,68 @@ public class BigQueryIO {
 
       @Override
       public void validate(PInput input) {
-        if (validate) {
-          BigQueryOptions bqOptions = input.getPipeline().getOptions().as(BigQueryOptions.class);
+        if (!validate) {
+          // Note that a table or query check can fail if the table or dataset are created by
+          // earlier stages of the pipeline or if a query depends on earlier stages of a pipeline.
+          // For these cases the withoutValidation method can be used to disable the check.
+          return;
+        }
 
-          String tempLocation = bqOptions.getTempLocation();
-          checkArgument(
-              !Strings.isNullOrEmpty(tempLocation),
-              "BigQueryIO.Read needs a GCS temp location to store temp files.");
-          if (testBigQueryServices == null) {
-            try {
-              GcsPath.fromUri(tempLocation);
-            } catch (IllegalArgumentException e) {
-              throw new IllegalArgumentException(
-                  String.format(
-                      "BigQuery temp location expected a valid 'gs://' path, but was given '%s'",
-                      tempLocation),
-                  e);
-            }
-          }
+        BigQueryOptions bqOptions = input.getPipeline().getOptions().as(BigQueryOptions.class);
 
-          TableReference table = getTableWithDefaultProject(bqOptions);
-          if (table == null && query == null) {
-            throw new IllegalStateException(
-                "Invalid BigQuery read operation, either table reference or query has to be set");
-          } else if (table != null && query != null) {
-            throw new IllegalStateException("Invalid BigQuery read operation. Specifies both a"
-                + " query and a table, only one of these should be provided");
-          } else if (table != null && flattenResults != null) {
-            throw new IllegalStateException("Invalid BigQuery read operation. Specifies a"
-                + " table with a result flattening preference, which is not configurable");
-          } else if (query != null && flattenResults == null) {
-            throw new IllegalStateException("Invalid BigQuery read operation. Specifies a"
-                + " query without a result flattening preference");
-          } else if (table != null && useLegacySql != null) {
-            throw new IllegalStateException("Invalid BigQuery read operation. Specifies a"
-                + " table with a SQL dialect preference, which is not configurable");
-          } else if (query != null && useLegacySql == null) {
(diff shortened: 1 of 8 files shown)


sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BigQueryIO.java
```diff
@@ -523,6 +549,12 @@ public void validate(PInput input) {
           } else if (query != null && flattenResults == null) {
             throw new IllegalStateException("Invalid BigQuery read operation. Specifies a"
                 + " query without a result flattening preference");
+          } else if (table != null && useLegacySql != null) {
```
peihe: This follows the pattern of flattenResults.

However, I think we can move the validation of useLegacySql to usingStandardSql(). For example, throws if users do BigQuery.Read.from(table).usingStandardSql().

Also, looks to me flattenResults is not configurable, and can be removed in a separate PR.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BigQueryIO.java
```diff
@@ -460,17 +467,23 @@ public Bound from(TableReference table) {
...
+            testBigQueryServices);
       }
 
       /**
        * Disable table validation.
        */
       public Bound withoutValidation() {
-        return new Bound(name, query, jsonTableRef, false, flattenResults, testBigQueryServices);
+        return new Bound(
+            name, query, jsonTableRef, false, flattenResults, useLegacySql, testBigQueryServices);
```
peihe: false /* validate */

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BigQueryIO.java
```diff
@@ -481,12 +494,25 @@ public Bound withoutValidation() {
        * from a table will cause an error during validation.
        */
       public Bound withoutResultFlattening() {
-        return new Bound(name, query, jsonTableRef, validate, false, testBigQueryServices);
+        return new Bound(
+            name, query, jsonTableRef, validate, false, useLegacySql, testBigQueryServices);
```
peihe: false /* flattenResults*/

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BigQueryIO.java
```diff
@@ -481,12 +494,25 @@ public Bound withoutValidation() {
...
+
+      /**
+       * Enables BigQuery's Standard SQL dialect when reading from a query.
+       *
+       * <p>Only valid when a query is used ({@link #fromQuery}). Setting this option when reading
+       * from a table will cause an error during validation.
+       */
+      public Bound usingStandardSql() {
+        return new Bound(
+            name, query, jsonTableRef, validate, flattenResults, false, testBigQueryServices);
```
peihe: false /* useLegacySql */

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BigQueryIO.java
peihe: I took my comment back, I see users could do BigQuery.Read.named().usingStandardSql().from(table) or even BigQuery.Read.fromQuery(query).usingStandardSql().from(table).

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BigQueryIO.java
```diff
@@ -534,16 +566,20 @@ public void validate(PInput input) {
...
           if (query != null) {
-            dryRunQuery(bqOptions, query);
+            dryRunQuery(bqOptions, query, useLegacySql);
           }
         }
       }
 
-      private static void dryRunQuery(BigQueryOptions options, String query) {
+      private static void dryRunQuery(
+          BigQueryOptions options, String query, @Nullable Boolean useLegacySql) {
```
peihe: useLegacySql should not be null for query

sdk/src/test/java/com/google/cloud/dataflow/sdk/io/BigQueryIOTest.java
```diff
@@ -575,6 +576,25 @@ public void testBuildSourceWithTableAndFlatten() {
...
+    thrown.expectMessage(
+        "Invalid BigQuery read operation. Specifies a"
+              + " table with a SQL dialect preference, which is not configurable");
+    p.apply(
+        BigQueryIO.Read.named("ReadMyTable")
+            .from("foo.com:project:somedataset.sometable")
+            .usingStandardSql());
+    p.run();
+  }
+
```
peihe: What is the plan to test the success case?

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BigQueryIO.java
dhalperi: No, it definitely can -- it's null if the user has not set it. Look ~10 lines up

sdk/src/test/java/com/google/cloud/dataflow/sdk/io/BigQueryIOTest.java
dhalperi: Not sure what you're asking.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BigQueryIO.java
dhalperi: You're right, we enforce this property in the constructor. PTAL

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/BigQueryIO.java
```diff
@@ -481,69 +495,93 @@ public Bound withoutValidation() {
...
+              "Invalid BigQueryIO.Read: Specifies a table with a SQL dialect"
+                  + " preference, which only applies to queries");
+
+          // Check for source table presence for early failure notification.
+
+          verifyDatasetPresence(bqOptions, table);
+          verifyTablePresence(bqOptions, table);
+        } else /* query != null */ {
+          checkState(flattenResults != null, "flattenResults should not be null if query is set");
+          checkState(useLegacySql != null, "useLegacySql should not be null if query is set");
```
peihe: checkNotNull for both?

dhalperi: I want to throw `IllegalStateException` not `NullPointerException`

PR #432, opened 2016-09-13
SDK: suppress expected deprecation warnings

dhalperi:
R: @tgroh


dhalperi:
Some of those surprise me as I did not see them locally (e.g., the DatastoreIOTest should be covered by the class-level suppression). I will merge this and then look again.

PR #438, opened 2016-09-20
Add Flatten with Duplicate Inputs Test

tgroh:
This test demonstrates that if a Flatten.PCollections transform has the
same PCollectionList in its input multiple times, it will duplicate the
input PCollection for each time it occurs within the list.

Backports https://github.com/apache/incubator-beam/pull/960


lukecwik:
LGTM, will merge after checks pass

PR #439, opened 2016-09-20
Use encoded Elements in the DirectPipelineRunner

### LINKED ISSUE #434: DataflowAssert serde cycle on actual values
Afaiu `CheckRelationAgainstExpected` from [DataflowAssert](https://github.com/GoogleCloudPlatform/DataflowJavaSDK/blob/master/sdk/src/main/java/com/google/cloud/dataflow/sdk/testing/DataflowAssert.java#L1022), it does serde cycle on expected values, while actual values coming from PCollection would not go through the same cycle. This may lead to inconsistency, for example in case of TableRow/TableRowJsonCoder/Jackson:
   
  * http://stackoverflow.com/questions/33622227/dataflow-mixing-integer-long-types
  * http://stackoverflow.com/questions/7316689/forcing-jackson-to-deserialize-to-specific-primitive-type

What do you think about enforcing serde cycle on PCollection before matching with `expected` values?

This issue came up in spotify/scio#252

tgroh:
This should fix #434

### DIFF
--- a/sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/DirectPipelineRunner.java
+++ b/sdk/src/main/java/com/google/cloud/dataflow/sdk/runners/DirectPipelineRunner.java
@@ -937,8 +937,8 @@ public class DirectPipelineRunner
     public <T> void setPCollectionValuesWithMetadata(
         PCollection<T> pc, List<ValueWithMetadata<T>> elements) {
       LOG.debug("Setting {} = {}", pc, elements);
-      ensurePCollectionEncodable(pc, elements);
-      setPValue(pc, elements);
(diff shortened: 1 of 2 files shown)


tgroh:
R: @kennknowles

kennknowles:
Travis failure looks real but unrelated to this change. I just looked at the last error; maybe something went sideways in the log somewhere?

tgroh:
Yeah, it looks like `BoundedReadFromUnboundedSourceTest` exploded (`#testTimeBound` in at least one of the cases)

tgroh:
@kennknowles 

reducing the read time in `BoundedReadFromUnboundedSourceTest` continues to demonstrate that it applies a limited read, but reduces the pressure on Travis. PTAL.

PR #443, opened 2016-09-23
Add the removed public API back for backward compatibility.

peihe:
R: @lukecwik 
CC: @dhalperi

### DIFF
--- a/sdk/src/main/java/com/google/cloud/dataflow/sdk/util/BigQueryTableRowIterator.java
+++ b/sdk/src/main/java/com/google/cloud/dataflow/sdk/util/BigQueryTableRowIterator.java
@@ -121,8 +121,20 @@ public class BigQueryTableRowIterator implements AutoCloseable {
 
   /**
    * Constructs a {@code BigQueryTableRowIterator} that reads from the results of executing the
-   * specified query in the specified project.
+   * specified query in the specified project with useLegacySql set to True.
+   *
+   * @deprecated use {@link #fromQuery(String, String, Bigquery, Boolean, Boolean)}.
    */
+  @Deprecated
+  public static BigQueryTableRowIterator fromQuery(
+      String query, String projectId, Bigquery client, @Nullable Boolean flattenResults) {
+    return fromQuery(query, projectId, client, flattenResults, Boolean.TRUE);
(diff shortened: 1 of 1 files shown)


sdk/src/main/java/com/google/cloud/dataflow/sdk/util/BigQueryTableRowIterator.java
```diff
@@ -121,8 +121,20 @@ public static BigQueryTableRowIterator fromTable(TableReference ref, Bigquery cl
 
   /**
    * Constructs a {@code BigQueryTableRowIterator} that reads from the results of executing the
-   * specified query in the specified project.
+   * specified query in the specified project with useLegacySql set to True.
+   *
+   * @deprecated use {@link #fromQuery(String, String, Bigquery, Boolean, Boolean)}.
    */
+  @Deprecated
+  public static BigQueryTableRowIterator fromQuery(
```
dhalperi: This is in the `util` package -- is it really public API?

peihe: We can make none backward compatible changes in here.
However, I think users can still choose to use it since it is a public class. (Our integration tests use it)

I think maintaining this deprecated method has very little overheads, and might be worth to have it.
I am okay with either way.

sdk/src/main/java/com/google/cloud/dataflow/sdk/util/BigQueryTableRowIterator.java
```diff
@@ -121,8 +121,20 @@ public static BigQueryTableRowIterator fromTable(TableReference ref, Bigquery cl
...
+      String query, String projectId, Bigquery client, @Nullable Boolean flattenResults) {
+    return fromQuery(query, projectId, client, flattenResults, Boolean.TRUE);
```
dhalperi: why not `null` so the default is in only one place?

Also, add `/* useLegacySql */` to the constant either way.

peihe: done

PR #452, opened 2016-09-29
BoundedReadFromUnboundedSourceTest: fixes

dhalperi:
1) Tests actually require non-splitting source (assert that the Key is 0).
   Use withoutSplitting.

2) Indentation and unused code (expectedOutput).

3) For these time-based tests, runners cannot guarantee we'll produce
   at least 3 elements, but we can expect there to be at least 1.

These appear in Beam but were not backported:

https://github.com/apache/incubator-beam/commit/fecd0362f5c1710c7fee0b2a03cc8549d1b8f75a
https://github.com/apache/incubator-beam/commit/ee7272fe77d78240c6f8e0b79cdc93b460171b8c


lukecwik:
LGTM
Did the diffs, only substantial difference remaining is that in Apache Beam when refactoring the Direct runner out, we removed the finalize tracker support.

lukecwik:
Merging

PR #453, opened 2016-09-29
Fix the NPE when getting reference table location in BigQueryQuerySource

peihe:


sdk/src/test/java/com/google/cloud/dataflow/sdk/io/BigQueryIOTest.java
```diff
@@ -1214,6 +1214,84 @@ public void testBigQueryQuerySourceInitSplit() throws Exception {
   }
 
   @Test
+  public void testBigQueryNoTableQuerySourceInitSplit() throws Exception {
```
peihe: This test is copied from above and removed the table returned from the mock.

peihe:
R: @dhalperi

sdk/src/test/java/com/google/cloud/dataflow/sdk/io/BigQueryIOTest.java
dhalperi: Per discussion, we really need less code duplication in here. We're going to rewrite this code post-1.8 release.

PR #475, opened 2016-10-31
Use a NavigableSet Instead of a PriorityQueue in WatermarkManager

tgroh:
This removes an O(n) call to remove, replacing it with an O(log(n))
call. This significantly improves scaling behavior of the DirectRunner

Backports https://github.com/apache/incubator-beam/pull/1202


kennknowles:
LGTM. Feel free to self-merge after fixing checkstyle. Might just want to get Travis to pass.

PR #477, opened 2016-11-01
Makes UnionCoder public

jkff:
This is useful because users may need to construct a CoGbkResultCoder, and to construct a CoGbkResult Coder you need to construct a UnionCoder. E.g. see http://stackoverflow.com/questions/40367387/how-do-i-setcoders-when-doing-an-end-to-end-test-when-my-coder-requires-a-kvcode

UnionCoder is already public in Apache Beam.

R: @lukecwik


lukecwik:
LGTM, merging

This mirrors the change to UnionCoder.java within https://github.com/apache/incubator-beam/commit/24bfca230d5db3cb75dd0e30093a10f7523c1238

PR #479, opened 2016-11-04
Add SortValues

mizitch:
A contrib module that provides a PTransform which performs
local(non-distributed) sorting. It will sort in memory until the buffer
is full, then flush to disk and use external sorting.

Consumes a PCollection of KVs from primary key to iterable of secondary
key and value KVs and sorts the iterables. Would probably be called
after a GroupByKey. Uses coders to convert secondary keys and values
into byte arrays and does a lexicographical comparison on the secondary
keys.

Uses Hadoop as an external sorting library.

Backport of https://github.com/apache/incubator-beam/pull/1199

### DIFF
--- a/contrib/sorter/AUTHORS.md
+++ b/contrib/sorter/AUTHORS.md
@@ -1,0 +1,5 @@
+# Authors of 'sorter' module
+
+The following is the official list of authors for copyright purposes of this community-contributed module.
+
+    Google Inc.
--- a/contrib/sorter/AUTHORS.md
+++ b/contrib/sorter/AUTHORS.md
@@ -1,5 +1,0 @@
-# Authors of 'sorter' module
-
-The following is the official list of authors for copyright purposes of this community-contributed module.
-
-    Google Inc.
--- a/contrib/sorter/README.md
+++ b/contrib/sorter/README.md
@@ -1,0 +1,23 @@
+#Sorter
+This module provides the SortValues transform, which takes a `PCollection<KV<K, Iterable<KV<K2, V>>>>` and produces a `PCollection<KV<K, Iterable<KV<K2, V>>>>` where, for each primary key `K` the paired `Iterable<KV<K2, V>>` has been sorted by the byte encoding of secondary key (`K2`). It will efficiently and scalably sort the iterables, even if they are large (do not fit in memory).
+
+##Caveats
+* This transform performs value-only sorting; the iterable accompanying each key is sorted, but *there is no relationship between different keys*, as Beam does not support any defined relationship between different elements in a PCollection.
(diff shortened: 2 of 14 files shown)


mizitch:
Note: Everything is a pretty much direct backport *except* for pom.xml. The beam and dataflow poms are different enough that it was easier to take the original pom.xml Marian had written for the earlier version of the sort library and manually make the necessary changes to dependencies and shading. So aside from the CR done on Marian's version of the file months ago, that file hasn't been code reviewed.

Ran mvn clean verify successfully on the sorter contrib module and the parent (looks like the parent pom doesn't do builds of the contrib modules in dataflow sdk).

contrib/sorter/AUTHORS.md
```diff
@@ -0,0 +1,5 @@
+# Authors of 'sorter' module
```
davorbonaci: Not needed unless there's somebody other than Google to add here.

contrib/sorter/README.md
```diff
@@ -0,0 +1,23 @@
+#Sorter
+This module provides the SortValues transform, which takes a `PCollection<KV<K, Iterable<KV<K2, V>>>>` and produces a `PCollection<KV<K, Iterable<KV<K2, V>>>>` where, for each primary key `K` the paired `Iterable<KV<K2, V>>` has been sorted by the byte encoding of secondary key (`K2`). It will efficiently and scalably sort the iterables, even if they are large (do not fit in memory).
+
+##Caveats
+* This transform performs value-only sorting; the iterable accompanying each key is sorted, but *there is no relationship between different keys*, as Beam does not support any defined relationship between different elements in a PCollection.
```
davorbonaci: Beam -> Dataflow

contrib/sorter/src/test/java/com/google/cloud/dataflow/contrib/sorter/SortValuesTest.java
```diff
@@ -0,0 +1,127 @@
...
+              KvMatcher.isKv(
+                  is("key2"),
+                  contains(
+                      KvMatcher.isKv(is("secondaryKey1"), is(100)),
+                      KvMatcher.isKv(is("secondaryKey2"), is(200))))));
+      return null;
+    }
+  }
+
+  /** Matcher for KVs. Forked from Beam's com.google.cloud.dataflow/sdk/TestUtils.java */
```
davorbonaci: Beam?

contrib/sorter/AUTHORS.md
mizitch: removed

contrib/sorter/README.md
mizitch: Realizing I should have done a case insensitive grep for beam in the first place... Have done that now and the only two instances were the ones you found. Fixed

contrib/sorter/src/test/java/com/google/cloud/dataflow/contrib/sorter/SortValuesTest.java
mizitch: Fixed

dhalperi:
LGTM, thanks.

PR #482, opened 2016-11-10
Backporting FileBasedSink changes from Beam

jkff:
Original PRs:
https://github.com/apache/incubator-beam/pull/1050
https://github.com/apache/incubator-beam/pull/1278

Changes simply copied over. No modifications.

R: @lukecwik

### DIFF
--- a/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/FileBasedSink.java
+++ b/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/FileBasedSink.java
@@ -31,20 +31,26 @@ import com.google.cloud.dataflow.sdk.util.IOChannelFactory;
 import com.google.cloud.dataflow.sdk.util.IOChannelUtils;
 import com.google.cloud.dataflow.sdk.util.MimeTypes;
 
+import org.joda.time.Instant;
+import org.joda.time.format.DateTimeFormat;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.File;
 import java.io.IOException;
 import java.io.Serializable;
 import java.nio.channels.WritableByteChannel;
 import java.nio.file.Files;
 import java.nio.file.NoSuchFileException;
+import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 /**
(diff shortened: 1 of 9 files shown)


sdk/src/main/java/com/google/cloud/dataflow/sdk/io/FileBasedSink.java
```diff
@@ -370,21 +390,18 @@ public void finalize(Iterable<FileResult> writerResults, PipelineOptions options
...
-    protected final void removeTemporaryFiles(PipelineOptions options) throws IOException {
-      String pattern = buildTemporaryFilename(baseTemporaryFilename, "*");
-      LOG.debug("Finding temporary bundle output files matching {}.", pattern);
-      FileOperations fileOperations = FileOperationsFactory.getFileOperations(pattern, options);
-      IOChannelFactory factory = IOChannelUtils.getFactory(pattern);
-      Collection<String> matches = factory.match(pattern);
-      LOG.debug("{} temporary files matched {}", matches.size(), pattern);
-      LOG.debug("Removing {} files.", matches.size());
-      fileOperations.remove(matches);
+    protected final void removeTemporaryFiles(List<String> knownFiles, PipelineOptions options)
```
dhalperi: not backwards-compatible

jkff: The rest of the current change is technically also not backward-compatible (e.g. the set of protected member variables has changed) for someone who was overriding FileBasedSink/FileBasedWriteOperation/... in a highly non-trivial way (e.g. trying to construct filenames manually using the protected members).

Note that, of the file-based sinks I've seen, none would be broken by this change, because they don't peek into these protected variables.

However, as for those that potentially do, I don't think there's a way to make the current change backward-compatible for them. There's a choice:

- Make the current change, with a (I think) very small chance of breaking somebody, but in return, fix deletion of files and data loss/duplication issues.
- Abandon the current change and say that the problem is only fixed in the Beam SDK
- Do something much weaker but backward-compatible instead: e.g., simply add code for removing all known files, but do not place temp files in a subdirectory. As an unfortunate side effect, this will cause a divergence in code and behavior between Dataflow and Beam SDKs.

Which would you prefer? Are there other options?

davorbonaci: #3 is the default choice, unless *very* strong guarantees can be made regarding #1.

davorbonaci: @lukecwik, you said it so nicely ;-)

jkff:
OK, I'm closing this PR in favor of #484.

PR #484, opened 2016-11-14
FileBasedSink removes temp files from successful bundles

jkff:
This is a weaker but backward-compatible version of #482.

R: @dhalperi @lukecwik

### DIFF
--- a/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/FileBasedSink.java
+++ b/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/FileBasedSink.java
@@ -44,7 +44,9 @@ import java.nio.file.StandardCopyOption;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
 /**
  * Abstract {@link Sink} for file-based output. An implementation of FileBasedSink writes file-based
@@ -312,7 +314,7 @@ public abstract class FileBasedSink<T> extends Sink<T> {
 
       // Optionally remove temporary files.
       if (temporaryFileRetention == TemporaryFileRetention.REMOVE) {
-        removeTemporaryFiles(options);
+        removeTemporaryFiles(files, options);
       }
     }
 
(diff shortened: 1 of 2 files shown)


sdk/src/test/java/com/google/cloud/dataflow/sdk/io/FileBasedSinkTest.java
```diff
@@ -248,20 +248,20 @@ private void testRemoveTemporaryFiles(int numFiles, String baseTemporaryFilename
     PipelineOptions options = PipelineOptionsFactory.create();
     SimpleSink.SimpleWriteOperation writeOp = buildWriteOperation(baseTemporaryFilename);
 
-    List<File> temporaryFiles = new ArrayList<>();
+    List<String> temporaryFiles = new ArrayList<>();
     List<File> outputFiles = new ArrayList<>();
     for (int i = 0; i < numFiles; i++) {
```
lukecwik: Please add a unit test that specifically passes in temporary files that would have never been matched.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/FileBasedSink.java
```diff
@@ -370,20 +372,42 @@ public void finalize(Iterable<FileResult> writerResults, PipelineOptions options
...
       LOG.debug("Finding temporary bundle output files matching {}.", pattern);
       FileOperations fileOperations = FileOperationsFactory.getFileOperations(pattern, options);
       IOChannelFactory factory = IOChannelUtils.getFactory(pattern);
       Collection<String> matches = factory.match(pattern);
-      LOG.debug("{} temporary files matched {}", matches.size(), pattern);
-      LOG.debug("Removing {} files.", matches.size());
+      Set<String> allMatches = new HashSet<>(matches);
+      allMatches.addAll(knownFiles);
+      LOG.debug(
+          "Removing {} temporary files matching {} ({} matched glob, {} additional known files)",
```
lukecwik: This log statement might be confusing since X != Y + Z because of the set union. Is there a better way to explain this?

jkff: Not sure what you mean: the last parameter of the log message, Z, is defined as X - Y.

lukecwik: Nvm

PR #488, opened 2016-11-26
Extending BigqueryTableInserter to take maximum batch size

KotaiVictor:
We ran into an issue in production recently where one of our workers was stalling on a particular hotkey. 

Looking further into the issue we figured out that the given our row sizes we were inserting somewhere between 5 and 20 elements per call. 

This coupled with our windowing size of around 5000 and the fixed number of 100 threads the BigQueryTableInserter uses to do inserts made it slow down. 

We thought it would be useful to be able to configure this.

Ideally, we'd be able to set the thread count for the Executor too. Give that it's static, you'll be able to do at most 100 inserts in parallel from a worker.

### DIFF
--- a/sdk/src/main/java/com/google/cloud/dataflow/sdk/util/BigQueryTableInserter.java
+++ b/sdk/src/main/java/com/google/cloud/dataflow/sdk/util/BigQueryTableInserter.java
@@ -63,7 +63,7 @@ public class BigQueryTableInserter {
   private static final Logger LOG = LoggerFactory.getLogger(BigQueryTableInserter.class);
 
   // Approximate amount of table data to upload per InsertAll request.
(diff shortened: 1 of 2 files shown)


KotaiVictor:
I signed it!

dhalperi:
Looks fine to me, but don't you want a way to plumb the option actually from the `BigQueryIO.Write`? Otherwise, this is unused code.

Can you please submit any upgrades to [Apache Beam](http://beam.incubator.apache.org/contribute/contribution-guide/), which will be the basis of Dataflow 2.0? That way they will not be lost in future.

dhalperi:
Hi @KotaiVictor,

Due to inactivity, I'm going to close this PR for now. However, if you are ready to discuss again please let me know.

Thanks!

PR #489, opened 2016-11-28
Use a retrying Dataflow Client for DataflowPipelineJob

dpmills:
Beam already does this; this will make DataflowPipelineJob.cancel() non-flaky.


kennknowles:
Looks good. I'll wait for tests before merging.

kennknowles:
Three out of four Travis CI builds passed, and the one that failed looked like it was a Maven Central infrastructure issue that led to a timeout. Nonetheless, since it was also the only piece of the test matrix using OpenJDK, I've restarted the test run rather than merge.

kennknowles:
The test that failed previously has now passed, along with all Linux-based tests. The Mac OSX entry of the test matrix was already green. So this is good and I am merging.

PR #490, opened 2016-12-01
Suppress spurious rebuilds due to missing package-info.class

kennknowles:
This is a backport of apache/incubator-beam#1205 and subsequent improvements. It also demonstrates that fiddling with `sdk.properties` does not affect rebuilds.

I have confirmed locally that builds are idempotent with both JDK8 and JDK7. Additional confirmation welcome.

R: @dhalperi


dhalperi:
Thanks!

PR #496, opened 2016-12-07
Templated jobs should use custom IO for BigQuery

sammcveety:
R: @dhalperi 

Using custom IO for this anticipates working with Beam, and supports better iteration by not requiring service cooperation for template execution.  The data cleanup regressions are suboptimal, but acceptable.


dhalperi:
Can you say more about why the Dataflow service should not want to override BigQueryIO in the case of a templated pipeline? Are we accepting the data cleanup regressions in that case?

sammcveety:
Done

PR #501, opened 2016-12-13
Backport https://github.com/apache/incubator-beam/pull/1230

sammcveety:
R: @dhalperi 

Pulled in changes for PubsubUnbounded*.java from https://github.com/apache/incubator-beam/pull/1504

### DIFF
--- a/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/PubsubIO.java
+++ b/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/PubsubIO.java
@@ -140,7 +140,7 @@ public class PubsubIO {
    * Populate common {@link DisplayData} between Pubsub source and sink.
    */
   private static void populateCommonDisplayData(DisplayData.Builder builder,
-      String timestampLabel, String idLabel, String topic) {
+      String timestampLabel, String idLabel, ValueProvider<PubsubTopic> topic) {
     builder
         .addIfNotNull(DisplayData.item("timestampLabel", timestampLabel)
             .withLabel("Timestamp Label Attribute"))
@@ -148,7 +148,9 @@ public class PubsubIO {
             .withLabel("ID Label Attribute"));
 
     if (topic != null) {
-      builder.add(DisplayData.item("topic", topic)
+      String topicString = topic.isAccessible() ? topic.get().asPath()
+          : topic.toString();
+      builder.add(DisplayData.item("topic", topicString)
           .withLabel("Pubsub Topic"));
     }
   }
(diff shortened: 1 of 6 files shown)


sammcveety:
Rebased

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/PubsubIO.java
```diff
@@ -666,8 +754,11 @@ private Bound(String name, PubsubSubscription subscription, PubsubTopic topic,
 
       @Override
       public void populateDisplayData(DisplayData.Builder builder) {
-        super.populateDisplayData(builder);
-        populateCommonDisplayData(builder, timestampLabel, idLabel, topic);
+        String topicString =
+            topic == null ? null
+            : topic.isAccessible() ? topic.get().asPath()
+            : topic.toString();
+        populateCommonDisplayData(builder, timestampLabel, idLabel, topicString);
```
dhalperi: Shouldn't we pass the VP into `populateCommonDisplayData` and avoid the logic here?

sdk/src/test/java/com/google/cloud/dataflow/sdk/io/PubsubIOTest.java
```diff
@@ -70,29 +73,13 @@ public void testTopicValidationSuccess() throws Exception {
   }
 
   @Test
-  public void testTopicValidationBadCharacter() throws Exception {
```
dhalperi: Not sure I like this change. Given an SVP, we should be able to catch these bad names, shouldn't we?

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/PubsubIO.java
sammcveety: Yes, yes we should.

sdk/src/test/java/com/google/cloud/dataflow/sdk/io/PubsubIOTest.java
sammcveety: Added it back, agreed.

sdk/src/main/java/com/google/cloud/dataflow/sdk/io/PubsubIO.java
```diff
@@ -140,15 +140,17 @@ private static void validatePubsubName(String name) {
...
         .addIfNotNull(DisplayData.item("timestampLabel", timestampLabel)
             .withLabel("Timestamp Label Attribute"))
         .addIfNotNull(DisplayData.item("idLabel", idLabel)
             .withLabel("ID Label Attribute"));
 
     if (topic != null) {
-      builder.add(DisplayData.item("topic", topic)
+      String topicString = topic.isAccessible() ? topic.get().asPath()
+          : topic.toString();
+      builder.add(DisplayData.item("topic", topicString)
```
dhalperi: Answered: it only accepts VP<KnownType> https://github.com/apache/incubator-beam/blob/master/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/display/DisplayData.java#L880

PR #502, opened 2016-12-13
checkstyle: improve Javadoc checking

dhalperi:
This is a spiritual backport of https://github.com/apache/incubator-beam/pull/1060

Same checkstyle.xml changes, very similar fixes but recreated because of code divergence


dhalperi:
The reason to do this is to make sanity-checking backports easier.

PR #509, opened 2016-12-15
Backport: [BEAM-1160] Fail to split in FileBasedSource if filePattern expands to empty

swegner:
Typically, input file patterns are validated during Pipeline
construction, but standard Read transforms include an option to disable
validation. This is generally useful but can lead to cases where a
Pipeline executes successfully with empty inputs.

This changes the behavior to fail execution on empty file-based inputs
even when validation is disabled.


swegner:
Optimistic backport assuming https://github.com/apache/incubator-beam/pull/1621 is ready to go.

R: @dhalperi

PR #510, opened 2016-12-15
Sgmc/backport 1400

sammcveety:
FYI @reuvenlax


dhalperi:
Sounds like this is being abandoned in favor of #514

PR #519, opened 2017-01-04
Add Dataflow, version 2.0.0-beta1, as a distribution of Apache Beam

davorbonaci:


sdk/src/main/resources/org/apache/beam/runners/dataflow/dataflow.properties
```diff
@@ -0,0 +1,4 @@
+environment.major.version=6
+
+worker.image.batch=dataflow.gcr.io/v1beta3/java-batch:${project.version}-20160103
```
dhalperi: these need to be `beam-java-batch`. My test job failed.

davorbonaci: Done.

PR #521, opened 2017-01-06
Add NOTICE to archetypes and enforce they match the root NOTICE

dhalperi:


dhalperi:
Alternate tried:

- Instead of copying `NOTICE` in the root resources, use `../../../../../../NOTICE` to try to pick it up from root. Did not work, but did not throw an error either.

.travis.yml
```diff
@@ -44,7 +44,12 @@ before_install:
   - if [ "$TRAVIS_OS_NAME" == "linux" ]; then jdk_switcher use "$CUSTOM_JDK"; fi
 
 install:
-  - travis_retry mvn install clean -U -DskipTests=true
+  - travis_retry mvn install -U -DskipTests=true
+  - mvn clean
```
davorbonaci: `mvn install clean` above, instead of two separate commands?

dhalperi: I don't love overly complicated Maven command lines, but reverted.

PR #531, opened 2017-01-11
[BEAM-359] Treat erased type variables as non-deterministic in AvroCoder

ehlyzov:
This is a back port of a BEAM-359. It allows to create a coder for generics in a case if the schema for the concrete type parameter is provided. The created coder will be treated as non-deterministic.

For more information look at: 
https://issues.apache.org/jira/browse/BEAM-359
https://github.com/apache/beam/commit/9c8dc4fe920618253b425f0d998f8d63552ec358
http://stackoverflow.com/questions/41381648/serializing-generic-class-with-provided-schema-using-avrocoder


dhalperi:
Thanks for this contribution, merging. This change will be available in `master` once merged.

If we make another release in the 1.x series, it will be included there, but this is not yet planned.

Issue #532, opened 2017-01-12
Override default `ACK_TIMEOUT_SEC` when creating random subs for PubsubIO


codingtwinky:
When [creating random subscription](https://github.com/GoogleCloudPlatform/DataflowJavaSDK/blob/bdbe05ec68e45a3082022649b8b5d68d6a56bfa9/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/PubsubIO.java#L860-L861) for PubsubIO, code uses hardcoded constant value and there is no way to override this value.

```scala 
subscriptionPath = pubsubClient.createRandomSubscription(projectPath, topicPath, ACK_TIMEOUT_SEC);
```

It would be nice to configurable option to override this value.

PR #533, opened 2017-01-12
Add a way to configure ackDeadlineSeconds

codingtwinky:
fixes: #532 

Does no validation of `ackDeadlineSeconds`, I believe acceptable range is t >= 0, and let the exception to be thrown at the random subscription creation time.

### DIFF
--- a/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/PubsubIO.java
+++ b/sdk/src/main/java/com/google/cloud/dataflow/sdk/io/PubsubIO.java
@@ -480,6 +480,13 @@ public class PubsubIO {
     }
 
     /**
+     * Set ackDeadlineSeconds for randomly created subscription when subscription is not specified.
+     */
+    public static Bound<String> ackDeadlineSeconds(int ackDeadlineSeconds) {
+      return new Bound<>(DEFAULT_PUBSUB_CODER).ackDeadlineSeconds(ackDeadlineSeconds);
+    }
+
+    /**
      * Creates and returns a transform reading from Cloud Pub/Sub where record timestamps are
      * expected to be provided as Pub/Sub message attributes. The {@code timestampLabel}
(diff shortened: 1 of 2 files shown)


codingtwinky:
I signed it!

dhalperi:
Hi @CodingTwinky -- this will not actually affect behavior when reading from Pub/Sub on Google Cloud Dataflow. Is this intended to change testing behavior using the `InProcessPipelineRunner`?

codingtwinky:
My intention was to add a way for user to configure `ackDeadlineSeconds` when no subscription is defined and random one is created via dataflow.  Am I missing something?

dhalperi:
@codingtwinky the code that creates the temporary subscription in the Java SDK is only used when running with the `InProcessPipelineRunner`. the `DataflowPipelineRunner` uses different code.

Can you say more information about why you want to change the ack deadline? You might want to reach out to the Dataflow support team: https://cloud.google.com/dataflow/support if you are seeing issues when running on the `DataflowPipelineRunner` that you believe are attributable to this parameter.

codingtwinky:
Reason why we were experimenting with various ack deadline was and experimentation to see dataflow's behavior during high backlog situation, which we've been having hard time with.  We were able to test by creating subscription manually but if we could configure ack deadline for throwaway subscription we can create on dataflow creation.

reuvenlax:
Dataflow should automatically extend acks if processing takes longer than the deadline. What exactly were the effects you saw by changing the deadline on the subscription?

codingtwinky:
For our use case, once dataflow failed to keep up with pubsub's publish rate, ~18k/sec, it gets into weird "ack deadline passed" death spiral and only way to recover is to pubsub's publish rate to go down.  

Increasing pubsub deadline ack timeout was an attempt at delaying when this death spiral happens.

Considering I can manually create a pubsub subscription with desired configuration I will close this issue but it would be nice to be able to configure this.

PR #534, opened 2017-01-13
Fix HadoopFileSource’s split size estimate

igorbernstein2:


contrib/hadoop/src/main/java/com/google/cloud/dataflow/contrib/hadoop/HadoopFileSource.java
```diff
@@ -239,12 +239,18 @@ public void validate() {
...
+      }
+
       Job job = Job.getInstance(); // new instance
       for (FileStatus st : listStatus(createFormat(job), job)) {
         size += st.getLen();
       }
     } catch (IOException | NoSuchMethodException | InvocationTargetException
-        | IllegalAccessException | InstantiationException e) {
+        | IllegalAccessException | InstantiationException | InterruptedException e) {
       // ignore, and return 0
```
dhalperi: This change does not look right. At the very least, you should restore the thread interrupted state when catching an InterruptedException.

Maybe move that to its own `catch` block and handle specifically?

igorbernstein2:
Fixed handling of InterruptedException. However I don't think this fix applies to beam. I don't think HadoopFileSource ever made it to apache beam.

dhalperi:
It got renamed to `HDFSFileSource`: https://github.com/apache/beam/blob/master/sdks/java/io/hdfs/src/main/java/org/apache/beam/sdk/io/hdfs/HDFSFileSource.java

igorbernstein2:
Thanks for pointing it out. Will submit a PR there as well.

PR #546, opened 2017-02-02
Upgrade to Apache Beam, version 0.5.0

davorbonaci:
This is based on the release candidate #2, currenly being considered as the release.

Lots of manual work, double-checking would be beneficial. This goes to `v2` branch, before an RC is cut. Some stuff comes later (archetype versioning, container image).


dhalperi:
Travis linux build passed, merging. OSX queue is stuck as per usual.

PR #547, opened 2017-02-02
Fix InProcessPipelineRunner to handle a null subscription

sammcveety:
R: @tgroh


sammcveety:
For verifying the subscription, is the least-bad way I can see is to make an additional factory method that includes a verifyCreateSubscription toggle.  Anything nicer than that?

dhalperi:
better link: https://github.com/apache/beam/pull/1817

sammcveety:
I think this is a different issue (not present in Beam), made obsolete by
the new structure of the function in question.

tgroh:
I don't think there's anything nicer than that, but I'm not particularly bothered (given that it's a `PubsubTestClientFactory`)

sammcveety:
Done, please take another look.

PR #558, opened 2017-03-17
Upgrade to Apache Beam, version 0.6.0

jasonkuster:
Lots of manual changes -- hopefully some which will make this process easier in the future.

Verified via mvn clean install, then generating, and then running WordCount in both examples and examples-java8 and StarterPipeline from starter.

R: @davorbonaci @aaltay


jasonkuster:
Note: As-is, will not pass -- need to update container image to 2.0.0-beta3 (omitting -20170202); 20170202 container is missing changes from Beam 0.6.0.

maven-archetypes/pom.xml
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
```
davorbonaci: Google header goes here.

maven-archetypes/starter/pom.xml
```diff
@@ -41,5 +42,44 @@
         <version>2.4</version>
       </extension>
     </extensions>
+
+    <pluginManagement>
+      <plugins>
+        <plugin>
+          <artifactId>maven-archetype-plugin</artifactId>
```
davorbonaci: Note to self: I wonder why this is not in the parent?

maven-archetypes/pom.xml
jasonkuster: Oops, fixed.

PR #562, opened 2017-04-04
Fixed documentation

anilmuppalla:


dhalperi:
Valid corporate CLA.

dhalperi:
Thanks for this nice fixup.

dhalperi:
Btw, would you also send a PR to Apache Beam, which is the basis of Dataflow 2? Looks like we have the same bug there: https://github.com/apache/beam/blob/master/sdks/java/core/src/main/java/org/apache/beam/sdk/transforms/windowing/AfterWatermark.java

anilmuppalla:
done. https://github.com/apache/beam/pull/2444

PR #566, opened 2017-04-28
Archetypes: version management at the top-level outside archetype source code

davorbonaci:
`Starter` reference intentionally not touched. Opinions on this?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -109,6 +109,11 @@
     <dataflow.container_version>${project.version}-20170317</dataflow.container_version>
     <dataflow.environment_major_version>6</dataflow.environment_major_version>
 
+    <!-- Archetype version propagated automatically to underlying archetype modules.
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -108,6 +108,8 @@
     <dataflow.release_name>Google Cloud Dataflow SDK for Java</dataflow.release_name>
     <dataflow.container_version>${project.version}-20170317</dataflow.container_version>
     <dataflow.environment_major_version>6</dataflow.environment_major_version>
+
+    <archetype.sdk_version_dependency>2.0.0-beta4-SNAPSHOT</archetype.sdk_version_dependency>
```
dhalperi: The set of things that are valid here and why deserves a comment.

davorbonaci: Done.

PR #574, opened 2017-05-23
Update versioning after cutting 2.0.0 release branch.

jasonkuster:


jasonkuster:
R: @davorbonaci @dhalperi

pom.xml
```diff
@@ -106,15 +106,15 @@
...
+    <dataflow.container_version>${project.version}-20170517</dataflow.container_version>
     <dataflow.environment_major_version>6</dataflow.environment_major_version>
 
     <!-- Archetype version propagated automatically to underlying archetype modules.
          Example values are: ${project.version}, 1.2.3-SNAPSHOT, [1.0.0,2),
                              [${project.version}, 2), etc.
          Starter archetype reference copy isn't affected by this value; a manual
          change is expected there. -->
-    <archetype.sdk_version_dependency>2.0.0-beta4-SNAPSHOT</archetype.sdk_version_dependency>
+    <archetype.sdk_version_dependency>2.1.0-SNAPSHOT</archetype.sdk_version_dependency>
```
jasonkuster: Is there a reason this is not (as cited in the comment) `${project.version}` here and on the release branch? It seems like setting it thus would obviate the need for this manual change.

pom.xml
```diff
@@ -106,15 +106,15 @@
...
 
     <dataflow.release_name>Google Cloud Dataflow SDK for Java</dataflow.release_name>
-    <dataflow.container_version>${project.version}-20170317</dataflow.container_version>
+    <dataflow.container_version>${project.version}-20170517</dataflow.container_version>
     <dataflow.environment_major_version>6</dataflow.environment_major_version>
 
     <!-- Archetype version propagated automatically to underlying archetype modules.
          Example values are: ${project.version}, 1.2.3-SNAPSHOT, [1.0.0,2),
                              [${project.version}, 2), etc.
          Starter archetype reference copy isn't affected by this value; a manual
```
jasonkuster: Starter archetype reference copy is pinned to `@project.version@`. Is this comment no longer accurate?

pom.xml
jasonkuster: Alternately, why are archetypes not just controlled by `@project.version@` as the starter archetype reference is? Under what circumstances will this version differ from `project.version`?

dhalperi: This was an explicit choice by @davorbonaci as a sanity check that things aren't working by accident.

pom.xml
dhalperi: *sigh* Maybe? @davorbonaci

dhalperi: Yes, this happened in https://github.com/GoogleCloudPlatform/DataflowJavaSDK/pull/569 and is at odds with both the comment and the value in the root pom and archetype resources pom.

Maybe just delete the variable and manage everything on `project.version`?

davorbonaci:
(lgtm either way -- feel free to change how the reference copy works, if you feel that's appropriate.)

jasonkuster:
We're going to pin to `@project.version@`.

PR #599, opened 2017-08-30
Update release versions in 2.1.0 branch.

wcn3:
Use released versions of artifacts.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -106,7 +106,7 @@
     <beam.version>2.1.0</beam.version>
 
     <dataflow.release_name>Google Cloud Dataflow SDK for Java</dataflow.release_name>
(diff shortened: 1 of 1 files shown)


wcn3:
I have the missing artifacts in my local .m2/repository as a result of building other parts of this release. The Travis CI doesn't, and tries to fetch from Sonatype. The process I'm following has the Sonatype push after this step, so there is a process deficiency to fix here.

wcn3:
The versioning update of this file can't happen until the artifacts have been pushed to official repos. I've reverted that portion of the change, and will have a new PR for that aspect once the artifacts are pushed.

lukecwik:
I still think you need to be able to update the version to 2.1.0 from 2.1.0-SNAPSHOT but can be in followup change.

PR #600, opened 2017-08-30
Update snapshot version of containers.

wcn3:


pom.xml
```diff
@@ -106,7 +106,7 @@
     <beam.version>2.1.0</beam.version>
```
aaltay: Should this be updated to `2.2.0-SNAPSHOT` ?

wcn3: Yes, you're right. I was confused about which version identifier to bump per the directions I was following.

wcn3: Actually, I don't believe that to be the case now.

I'm trying to replicate the effect of this comparable PR. 

https://github.com/GoogleCloudPlatform/DataflowJavaSDK/commit/f0ccc77f8f255e0b3cc5b2d570ce43e347d7dabe

The version dependency for the SDK should reflect the snapshot for the future release. The POM has been refactored since that change, so the SDK version is derived from project.version, which is properly updated by the release plugin. When I run 'mvn help:effective-pom', I see the updated version identifier matching my expectations.

wcn3:
This pull request has gotten broken because of changes I'd made to my repo. I didn't realize those changes would be reflected here. I'm closing this PR and will re-open a new one.

PR #608, opened 2017-10-05
Refactor versioning information for archetypes.

wcn3:
This makes versioning consistent with the Beam POMs. The top-level POM
contains the version numbers, and versions are plumbed into the
archetypes via properties.


lukecwik:
Confirmed that versions match Apache Beam v2.1.1 version except executor version which was explicitly changed.

PR #616, opened 2017-11-14
Reduce Size of UnboundedReadDeduplicatorTest

tgroh:
Fixes slowdowns causing timeouts in certain environments.


tgroh:
The travis failure is on a single environment due to environmental configuration, so I'm not convinced that it's a meaningful signal (given the other three environments on the matrix all pass still)

PR #627, opened 2018-03-27
Upgrade to Apache Beam version 2.4.0

akedin:


examples/pom.xml
```diff
@@ -20,7 +20,7 @@
   <parent>
     <groupId>com.google.cloud.dataflow</groupId>
     <artifactId>google-cloud-dataflow-java-sdk-parent</artifactId>
-    <version>2.4.0-SNAPSHOT</version>
+    <version>2.4.0</version>
```
lukecwik: On master we should only ever use x.y.z-snapshot, the release branch is the only place where we should ever have x.y.z as the version.

akedin: Done. Will update the doc accordingly

